PR #1, opened 2026-02-07
ci: add test workflow

rgmz:
Add the test workflow from Gitleaks to do some basic validation for pushes/prs.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -146,6 +146,7 @@ func TestTranslateAllowlists(t *testing.T) {
 						MatchCondition: AllowlistMatchAnd,
 						Paths:          []*regexp.Regexp{regexp.MustCompile("^node_modules/.*")},
 						StopWords:      []string{"mock"},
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,39 @@
...
+          go install gotest.tools/gotestsum@latest
+
+      - name: Test
+        run: |
+          gotestsum --raw-command -- go test -json ./... --race
+
+      # TODO: re-enable once generation is fixed.
+      - name: Validate Config
+        if: false
+        run: go generate ./... && git diff --exit-code
```
rgmz: The config/generate logic still relies on the older `detect` package. I haven't grokked the new code to update that, yet.

zricethezav: I was cleaning house and wiped `config/generate`. We could re-add it or we could consider a different approach. I'm not sure what but the config generation may be something we want to revisit.

rgmz: Cool, I'll remove that section for now.

PR #3, opened 2026-02-10
codec optimization

zricethezav:
Profiling showed `codec` to be a bottleneck. Running a `dir` scan against the gitlab repo w/ `max-decode-depth=0` yielded a 3.5-4s scan time vs a 9-10s scan time w/ the default `max-decode-depth=5` set.

This PR is mostly claude code generated using opus 4.6

### DIFF
--- a/scan/pipeline_test.go
+++ b/scan/pipeline_test.go
@@ -1,0 +1,473 @@
+package scan
+
+import (
+	"context"
+	"sort"
+	"testing"
+
+	"github.com/betterleaks/betterleaks"
+	"github.com/betterleaks/betterleaks/config"
+	_ "github.com/betterleaks/betterleaks/sources/file"
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


scan/codec/decoder_test.go
```diff
@@ -142,3 +143,307 @@ func TestDecode(t *testing.T) {
...
+				wantStr: "U+0041",
+			},
+			{
+				name:    "multiple U+XXXX sequences",
+				input:   "U+0048 U+0069",
+				wantStr: "U+0048 U+0069",
+			},
+			{
+				name:    "single backslash escape sequence",
+				input:   `\u0048\u0069\u0021\u0021\u0021\u0021`,
```
rgmz, line 294: Do we need more mixed cases to validate that this works with more realistic inputs?

zricethezav: Good call. We should probably include this test https://github.com/gitleaks/gitleaks/blob/master/detect/detect_test.go#L35-L91

PR #4, opened 2026-02-12
parallelize git scanning for sick gainz

zricethezav:
TODO add description and include activity monitor charts showing `git log` being a slog

### DIFF
--- a/scan/pipeline_test.go
+++ b/scan/pipeline_test.go
@@ -1,0 +1,473 @@
+package scan
+
+import (
+	"context"
+	"sort"
+	"testing"
+
+	"github.com/betterleaks/betterleaks"
+	"github.com/betterleaks/betterleaks/config"
+	_ "github.com/betterleaks/betterleaks/sources/file"
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


sources/git/parallel.go
```diff
@@ -0,0 +1,272 @@
...
+	"github.com/betterleaks/betterleaks/config"
+	"github.com/betterleaks/betterleaks/logging"
+	"github.com/fatih/semgroup"
+	"github.com/gitleaks/go-gitdiff/gitdiff"
+	"golang.org/x/sync/errgroup"
+)
+
+// ParallelGit scans a git repo by running multiple `git log -p` processes
+// in parallel, each covering a distinct set of commits. Commit SHAs are
+// enumerated once via `git rev-list` and then partitioned across workers,
```
rgmz, line 22: Smart!

How does `git rev-list` work for things like dangling commits or hidden refs (not under `refs/head/` or `refs/tag`)? Is there any chance that the list of commits is different from what you'd get following `git log -p`?

zricethezav: doesn't include them but neither does `git log -p --all` so this is an issue for both git modes we'll have to tackle. Follow up PR

PR #8, opened 2026-02-13
CLI flag in favor of build tags for regexp engines

zricethezav:


regexp/regexp.go
```diff
@@ -0,0 +1,70 @@
...
+
+// SetEngine selects the regex engine used by subsequent MustCompile calls.
+func SetEngine(name string) {
+	switch name {
+	case "stdlib", "re2":
+		currentEngine = name
+	default:
+		panic("regexp: unknown engine: " + name)
+	}
+}
```
rgmz: Question: if someone is using this as a library, is it valid to change the engine mid-execution?

zricethezav: that's a great question... Maybe we should protect `SetEngine` w/ a `sync.Once`?

rgmz: A global `SetEngine` that  you can only call once feels wrong. 

Although `MustCompile` already wraps the engine, maybe that's safe enough? If you compile 3 regexes with stdlib and switch to gore2, calling `MatchString` would use the respective engine, right?

zricethezav: > A global SetEngine that you can only call once feels wrong.

Yea fair, it'd be a public function. 

> Although MustCompile already wraps the engine, maybe that's safe enough? If you compile 3 regexes with stdlib and switch to gore2, calling MatchString would use the respective engine, right?

Correct. I don't foresee this really being a problem people actually run into so maybe a comment would suffice.

PR #18, opened 2026-02-21
fix: gracefully handle invalid regex

rgmz:
Porting https://github.com/gitleaks/gitleaks/pull/1847

### DIFF
--- a/regexp/regexp.go
+++ b/regexp/regexp.go
@@ -58,6 +58,24 @@ func SetEngine(name string) {
 	}
 }
 
+// Compile parses a regular expression using the currently selected engine.
+// If successful, returns a [Regexp] object that can be used to match against text.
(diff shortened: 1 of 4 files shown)


regexp/regexp.go
```diff
@@ -58,6 +58,24 @@ func SetEngine(name string) {
 	}
 }
 
+// Compile parses a regular expression using the currently selected engine.
+// If successful, returns a [Regexp] object that can be used to match against text.
+func Compile(str string) (*Regexp, error) {
```
zricethezav, line 63: I'm on board with this approach

rgmz: Aye; this seemed like the best way, but I'm open to changing it.

bplaxco (approved):
LGTM!

PR #22, opened 2026-02-23
`--match-context` let's users define match context by lines and characters

zricethezav:
This PR adds a `MatchContext` field to the finding struct which controls the output for `MatchContext`. Useful for helping determine if something is a FP

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -88,7 +88,7 @@ func init() {
 	rootCmd.PersistentFlags().Bool("no-banner", false, "suppress banner")
 	rootCmd.PersistentFlags().StringSlice("enable-rule", []string{}, "only enable specific rules by id")
 	rootCmd.PersistentFlags().StringP("gitleaks-ignore-path", "i", ".", "path to .betterleaksignore or .gitleaksignore file or folder containing one")
-	rootCmd.PersistentFlags().String("match-context", "", "bytes of context before and after a match (e.g. 100)")
+	rootCmd.PersistentFlags().Int("match-context", 0, "bytes of context before and after a match (e.g. 100)")
 	rootCmd.PersistentFlags().Int("max-decode-depth", 5, "allow recursive decoding up to this depth")
 	rootCmd.PersistentFlags().Int("max-archive-depth", 0, "allow scanning into nested archives up to this depth (default \"0\", no archive traversal is done)")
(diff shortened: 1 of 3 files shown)


detect/context.go
```diff
@@ -0,0 +1,132 @@
...
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+// MatchContext holds the parsed --match-context configuration.
+type MatchContext struct {
+	Lines int // number of lines before/after the match
+	Chars int // max characters per context line (0 = unlimited)
+}
```
rgmz: How does `Chars != 0` work if you have some lines that exceed the limit? Is it desirable to lose context for that specific line?

I like NoseyParker's approach where they do a flat number of bytes before/after.

zricethezav: > IHow does Chars != 0 work if you have some lines that exceed the limit? Is it desirable to lose context for that specific line?

Yea you would loose context
Screenshot 2026-02-23 at 10 38 05 AM

> I like NoseyParker's approach where they do a flat number of bytes before/after.

You may be right about that. Simpler. Convinced me

cmd/root.go
```diff
@@ -88,6 +88,7 @@ func init() {
 	rootCmd.PersistentFlags().Bool("no-banner", false, "suppress banner")
 	rootCmd.PersistentFlags().StringSlice("enable-rule", []string{}, "only enable specific rules by id")
 	rootCmd.PersistentFlags().StringP("gitleaks-ignore-path", "i", ".", "path to .betterleaksignore or .gitleaksignore file or folder containing one")
+	rootCmd.PersistentFlags().String("match-context", "", "bytes of context before and after a match (e.g. 100)")
```
zricethezav: damn come on now claude

detect/context.go
```diff
@@ -0,0 +1,44 @@
+package detect
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+// ParseMatchContext parses a --match-context value (number of bytes).
+func ParseMatchContext(spec string) (int, error) {
```
zricethezav: trash

rgmz (approved):
A potential usability improvement is to trim whitespace, and then re-add a space/newline if there was one in the boundary between the match and the before/after context.

This is useful in scenarios where the context contains a large amount of whitespace, which is both wasteful and harder to read (imo). Up to you.

PR #23, opened 2026-02-25
folks, it's time for some ai rules

zricethezav:
Added some missing big players

### DIFF
--- a/cmd/generate/config/rules/deepseek.go
+++ b/cmd/generate/config/rules/deepseek.go
@@ -12,7 +12,7 @@ func DeepSeek() *config.Rule {
 		Description: "Detected a DeepSeek API Key, which may expose AI model access and associated usage to unauthorized parties.",
 		Regex:       utils.GenerateSemiGenericRegex([]string{"deepseek"}, `sk-[a-f0-9]{32}`, true),
 		Keywords:    []string{"deepseek"},
-		Entropy:     3.7,
+		Entropy:     3.5,
 	}
 
 	tps := utils.GenerateSampleSecrets("deepseek", "sk-"+secrets.NewSecret(utils.Hex("32")))
--- a/config/betterleaks.toml
+++ b/config/betterleaks.toml
@@ -424,7 +424,7 @@ keywords = ["deepgram"]
 id = "deepseek-api-key"
 description = "Detected a DeepSeek API Key, which may expose AI model access and associated usage to unauthorized parties."
 regex = '''(?i)[\w.-]{0,50}?(?:deepseek)(?:[ \t\w.-]{0,20})[\s'"]{0,3}(?:=|>|:{1,3}=|\|\||:|=>|\?=|,)[\x60'"\s=]{0,5}(sk-[a-f0-9]{32})(?:[\x60'"\s;]|\\[nr]|$)'''
-entropy = 3.7
+entropy = 3.5
 keywords = ["deepseek"]
 
 [[rules]]


cmd/generate/config/rules/cursor.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/betterleaks/betterleaks/cmd/generate/secrets"
+	"github.com/betterleaks/betterleaks/config"
+)
+
+func CursorAPIKey() *config.Rule {
+	r := config.Rule{
+		RuleID:      "cursor-api-key",
+		Description: "Detected a Cursor Integrations API Key, which may expose AI-assisted development services to unauthorized access.",
+		Regex:       utils.GenerateSemiGenericRegex([]string{"cursor"}, `key_[0-9a-f]{64}`, true),
+		Keywords:    []string{"cursor"},
```
rgmz, line 14: Are we thinking that `cursor` will be more selective than `key_`?

cmd/generate/config/rules/endorlabs.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/betterleaks/betterleaks/cmd/generate/config/utils"
+	"github.com/betterleaks/betterleaks/cmd/generate/secrets"
+	"github.com/betterleaks/betterleaks/config"
+)
+
+func EndorLabsAPIKey() *config.Rule {
+	r := config.Rule{
+		RuleID:      "endorlabs-api-key",
+		Description: "Detected an Endor Labs API Key, which may compromise supply chain security scanning and software composition analysis.",
+		Regex:       utils.GenerateSemiGenericRegex([]string{"endor(?:labs)?", "key"}, `endr\+[A-Za-z0-9-]{16}`, true),
```
rgmz, line 13: Something defined with just `endr+` and no key/secret will be missed entirely. Is it worth having separate rules?

cmd/generate/config/rules/gitea.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/betterleaks/betterleaks/cmd/generate/config/utils"
+	"github.com/betterleaks/betterleaks/cmd/generate/secrets"
+	"github.com/betterleaks/betterleaks/config"
+)
+
+func GiteaAccessToken() *config.Rule {
+	r := config.Rule{
+		RuleID:      "gitea-access-token",
+		Description: "Detected a Gitea Access Token, which may expose self-hosted Git repositories and associated code to unauthorized access.",
+		Regex:       utils.GenerateSemiGenericRegex([]string{"gitea"}, utils.Hex("40"), true),
```
rgmz: This will probably need to be tweaked, I foresee it being noisy. Commit SHAs + Gitea would naturally be in the same place.

cmd/generate/config/rules/cursor.go
zricethezav: That was my thinking. Definitely can't just rely on `key_`

cmd/generate/config/rules/endorlabs.go
zricethezav: If we did that we would need some `specificity` feature to promote more specific findings. 

3 rules:
- api-key (high specificity, `key` in the match)
- api-secret (high specificity, `secret' in the match)
- api (anything with `endr+`)

Should we introduce a `specificity` field?

cmd/generate/config/rules/gitea.go
zricethezav: Good call

cmd/generate/config/rules/endorlabs.go
zricethezav: I'll make a follow up issue for this

PR #24, opened 2026-02-25
rule defined validation... long time coming

zricethezav:
Secrets Validation defined in rules is something I wanted to ship in Gitleaks for a long time. Now, I'm finally able to that that! (assuming this gets merged). This PR was based on the work originally done in Gitleaks here: https://github.com/gitleaks/gitleaks/pull/1519. 

This PR needs a lot of work and is admittedly most written by claude opus 4.6 (with plenty of steering ofc).

Couple of these we'll want to discuss:
1. templating
2. extractors
3. validation types (right now we are just proposing http)
4. `rule.Check` vs `rule.Validate` (this is for making sure each rule is clear for takeoff but makes for an awkward name since now we have the concept of Secrets Validation...)

@rgmz 

...

let's see what greptile has to say about this one

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ Flags:
       --diagnostics-dir string        directory to store diagnostics output files when not using http mode (defaults to current directory)
       --enable-rule strings           only enable specific rules by id
       --exit-code int                 exit code when leaks have been encountered (default 1)
+      --experiments string            comma-separated list of experimental features to enable (e.g. "validation")
   -i, --gitleaks-ignore-path string   path to .betterleaksignore or .gitleaksignore file or folder containing one (default ".")
   -h, --help                          help for betterleaks
       --ignore-gitleaks-allow         ignore betterleaks:allow and gitleaks:allow comments
@@ -454,7 +455,10 @@ fragment = section of data gitleaks is looking at
 ```
 
 
-#### Secret Validation
+#### Secrets Validation
+⚠️ **Secrets Validation is an experimental feature and likely to change. Feel free to use play with it and tweak your configs but don't be sad when we change it... because we are exploring _new_ and _cool_ ideas.** To enable the validation feature you must set the experimental flag w/ `--experiments=validation`
+⚠️
+
 
 Betterleaks can automatically check whether a detected secret is live by firing an HTTP request defined in a `[rules.validate]` block. Validation runs asynchronously during the scan with a pool of 10 workers.
 


cmd/generate/config/rules/config.tmpl
rgmz, line 1: Wow, GitHub finally has syntax highlighting for these templates. 😻

cmd/root.go
```diff
@@ -94,6 +95,12 @@ func init() {
 	rootCmd.PersistentFlags().Int("timeout", 0, "set a timeout for gitleaks commands in seconds (default \"0\", no timeout is set)")
 	rootCmd.PersistentFlags().String("regexp-engine", "re2", "regex engine (stdlib, re2)")
 
+	// Validation flags
+	rootCmd.PersistentFlags().Bool("validate-only", false, "only report confirmed findings (suppresses unvalidated + invalid)")
+	rootCmd.PersistentFlags().Bool("no-validate", false, "skip all validation even if rules have validate blocks")
```
rgmz: These flags are too similar.  `validate-only` also doesn't give enough control on the output -- what if you want to see valid and error?

---

How do you feel about the approach I took:
- Use `validate` as the feature flag, with it being true by default. Then `validate-timeout`, `validate-ttl`, or other related flags group nicely.
    ```go
    	rootCmd.PersistentFlags().BoolVar(&validationEnabled, "validate", true, "whether to attempt to validate secrets")
    	rootCmd.PersistentFlags().DurationVar(&validationTTL, "validation-ttl", 30*time.Minute, "how long to cache successful validation results")
    ```
- Use a separate set of flags to control console output, with more fine-grained control (specific names don't matter, I'm talking about the concept):
  ```go
  	rootCmd.PersistentFlags().BoolVar(&outputEnabled, "output-findings", true, "whether to print findings in real-time during scanning")
  	rootCmd.PersistentFlags().StringVar(&outputFormat, "output-format", "detailed", "real-time output format: 'detailed' or 'compact'")
  	rootCmd.PersistentFlags().StringVar(&outputStatus, "output-status", "all", "filter findings by validation status: valid,revoked,invalid,error,skipped,unsupported,all")
  	rootCmd.PersistentFlags().BoolVar(&outputUniqueOnly, "output-unique-only", true, "only output first occurrence of each unique finding (by fingerprint)")
  	rootCmd.PersistentFlags().StringVar(&outputMinConfidence, "output-min-confidence", "low", "minimum confidence level to display: low,medium,high")
  ```

README.md
```diff
@@ -424,6 +452,90 @@ fragment = section of data gitleaks is looking at
...
 
+#### Secret Validation
+
+> **Experimental feature**. Still ironing out the kinks...
+
+Betterleaks can automatically check whether a detected secret is live by firing an HTTP request defined in a `[rules.validate]` block. Validation runs asynchronously during the scan.
+
+Each `[rules.validate]` block describes an HTTP request and an ordered list of **match clauses**. Clauses are evaluated top-to-bottom; the first clause whose conditions all pass determines the finding's status. This first-match-wins design lets a single rule distinguish `confirmed`, `revoked`, `invalid`, etc. secrets from the same API endpoint.
+
+Template placeholders using `{{ rule-id }}` syntax are supported in `url`, `body`, and `headers`. At runtime, these are replaced with the secret(s) captured by the referenced rule. For composite rules with multiple required parts, all combinations are tested (cartesian product), and a single `confirmed` match is enough.
```
rgmz: I know we started with this approach, but these days I'd prefer some sort of mapping or implicit variable (e.g., `value` or `secret` for single-part rules, an explicit mapping for multi-part rules).

That way, the GitHub validation could be static, for example.
```go
...
		Headers: map[string]string{
			"Authorization": "token {{ secret }}",
			"Accept":        "application/vnd.github+json",
		},
...
```

zricethezav: I don't think I'm following here,
```
		Headers: map[string]string{
			"Authorization": "token {{ secret }}",
			"Accept":        "application/vnd.github+json",
		},
```

We generate this which is equivalent? 
```toml 
headers = { Accept = "application/vnd.github+json", Authorization = "token {{ github-fine-grained-pat }}" }
```

rgmz: Interpolation _for all rules_ would use a default generic variable (such as `secret`) instead of being tied to the rule-id. 

The rule's validation refer to it's own ID feels awkward. Maybe that's not a problem; the implicit linking to other rules (if we do it) would also be nice, I think.

zricethezav: how about this

First we update all the rules with named capture groups which can be referenced by the rule its defined in _or_ other rules via `{rule_id}.{named_capture}`. Default name for a rule would be `secret` if no named group is set.

zricethezav: named captures is something we gotta do i think

cmd/root.go
```diff
@@ -94,6 +95,13 @@ func init() {
 	rootCmd.PersistentFlags().Int("timeout", 0, "set a timeout for gitleaks commands in seconds (default \"0\", no timeout is set)")
 	rootCmd.PersistentFlags().String("regexp-engine", "re2", "regex engine (stdlib, re2)")
 
+	// Validation flags
+	rootCmd.PersistentFlags().Bool("validate", true, "enable validation of findings against live APIs")
+	rootCmd.PersistentFlags().String("output-status", "", "comma-separated list of validation statuses to include (e.g. confirmed,revoked)")
```
zricethezav: fail, this needs to be named something like `validation-statuses`

config/rule.go
```diff
@@ -58,6 +58,10 @@ type Rule struct {
 	// to measure how "rare" or non-natural-language a string is. Strings that
 	// tokenize efficiently (i.e., common words/phrases) are filtered out.
 	TokenEfficiency bool
+
+	// Validate describes an HTTP request to fire to determine whether
+	// a detected secret is live (confirmed) or stale/invalid.
+	Validate *Validation
```
zricethezav: `Validate` `Validation` and `Validator`... smdh, there must be a better way

config/rule.go
```diff
@@ -66,8 +70,8 @@ type Required struct {
 	WithinColumns *int
 }
 
-// Validate guards against common misconfigurations.
-func (r *Rule) Validate() error {
+// Check guards against common misconfigurations.
+func (r *Rule) Check() error {
```
zricethezav: Rename `CheckForMisconfiguration`

config/validate.go
```diff
@@ -0,0 +1,297 @@
...
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/betterleaks/betterleaks/logging"
+	"github.com/osteele/liquid"
+	"github.com/tidwall/gjson"
+)
+
+var validMethods = map[string]struct{}{
```
zricethezav, line 16: rename `validHTTPMethods`

zricethezav: We'll eventually handle more than just HTTP

config/validate.go
```diff
@@ -0,0 +1,297 @@
...
+	ValidationTypeHTTP ValidationType = "http"
+)
+
+// Validation describes a request to fire against a live API and a list of
+// match clauses to evaluate the response, determining a finding's status.
+type Validation struct {
+	Type    ValidationType
+	Method  string
+	URL     string
+	Headers map[string]string
```
zricethezav, line 37: How do we feel about having Type specific fields here like Headers? Probably a more elegant way to handle it but I figure as we add validation types we just add whatever we need here as we need it.

config/validate.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	return true
+}
+
+// matchJSONAssertion checks a GJSON result against an expected value:
+//   - "!empty": path must exist and be non-empty
+//   - []any: actual value must match one of the list entries
+//   - scalar: exact match via string comparison
+func matchJSONAssertion(result gjson.Result, expected any) bool {
```
zricethezav, line 240: eventually we could implement this as CEL instead of relying on a mini DSL

config/validate.go
rgmz: My gut was that we should make Validation an interface with type-specific structs ([for example](https://github.com/rgmz/lagotto/blob/8036c08fcfc1a437ebb0d51296f87702e37b058c/core/rule/validate/strategy.go#L10-L13)).

Interestingly, ProjectDiscovery uses one struct with each possible type as a field. :thinking: 
https://github.com/projectdiscovery/nuclei/blob/9142eae3a7dd9cb0c3c6113648212f39156f7088/pkg/templates/templates.go#L78

zricethezav: Sounds like the right place for an small interface like you've define. I like it

zricethezav: Holding off until we introduce a new http strat. Baby steps

cmd/root.go
zricethezav: @rgmz 
```golang
	// Validation flags
	rootCmd.PersistentFlags().Bool("validate", true, "enable validation of findings against live APIs")
	rootCmd.PersistentFlags().String("validation-status", "", "comma-separated list of validation statuses to include: confirmed, invalid, revoked, error, unknown")
	rootCmd.PersistentFlags().Duration("validate-timeout", 10*time.Second, "per-request timeout for validation")
	rootCmd.PersistentFlags().Bool("full-validation-response", false, "include full HTTP response body on validated findings")
```

Thoughts, concerns?

detect/detect.go
```diff
@@ -869,6 +898,58 @@ func (d *Detector) Findings() []report.Finding {
...
+// StartValidation spins up a pool of workers that validate findings asynchronously.
+// Findings whose rule has a [rules.validate] block will be sent to the pool from AddFinding.
+// Call DrainValidation after the scan completes to wait for all workers and collect results.
+func (d *Detector) StartValidation(ctx context.Context, v *validate.Validator, workers int) {
+	d.validator = v
+	d.validateCtx = ctx
+	d.validateCh = make(chan report.Finding, workers*2)
+	d.validated = make([]report.Finding, 0)
+
+	for range workers {
```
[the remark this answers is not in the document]
zricethezav: wrong

zricethezav:
> 8:21PM FTL Failed to Validate. True positive was not detected by regex. regex="(?i)[\\w.-]{0,50}?(?:wandb|weightsandbiases)(?:[ \\t\\w.-]{0,20})[\\s'\"]{0,3}(?:=|>|:{1,3}=|\\|\\||:|=>|\\?=|,)[\\x60'\"\\s=]{0,5}([a-f0-9]{40})(?:[\\x60'\"\\s;]|\\\\[nr]|$)" rule=weights-and-biases-api-key value="<wandb_api_keyToken>\n    2eee0ebe2f8d10900f8191f109f04db3e4467e08\n</wandb_api_keyToken>"
exit status 1

I am so trash. https://github.com/betterleaks/betterleaks/issues/27 please dear community

README.md
```diff
@@ -99,11 +100,16 @@ Flags:
...
+      --validation-status string        comma-separated list of validation statuses to include: confirmed, invalid, revoked, error, unknown, none (none = rules without validation)
       --redact uint[=100]             redact secrets from logs and stdout. To redact only parts of the secret just apply a percent value from 0..100. For example --redact=20 (default 100%)
   -f, --report-format string          output format (json, csv, junit, sarif, template)
   -r, --report-path string            report file
       --report-template string        template file used to generate the report (implies --report-format=template)
       --timeout int                   set a timeout for betterleaks commands in seconds (default "0", no timeout is set)
+      --validate                      enable validation of findings against live APIs (default true)
+      --validate-timeout duration     per-request timeout for validation (default 10s)
+      --validate-extract-empty         include empty values from extractors in output
+      --validation-full-response      include full HTTP response body on validated findings
```
rgmz: Is there a rule/convention for when something is `--validate-` versus `--validation-`?

README.md
```diff
@@ -299,6 +305,30 @@ id = "gitlab-pat"
...
+    url = "https://api.example.com/v1/verify"
+    headers = { Authorization = "Token {{ secret }}" }
+    extract = { user = "json:user.email", scopes = "header:X-OAuth-Scopes" }
+
+    # match is a first-match-wins list; the first clause whose conditions all
+    # pass determines the finding status.
+    match = [
+        { status = 200, json = { active = true }, result = "confirmed" },
+        { status = 401, result = "invalid" },
+        { result = "unknown" },
```
rgmz: Is this for demonstration purposes?

> Betterleaks will log a warning if the last clause is not a catch-all.

_Personally_, unmatched producing 'unknown' and logging a warning by default seems like a better UX. Manually defining a catch-all per rule can easily degrade into noise. 

https://github.com/rgmz/lagotto/blob/8036c08fcfc1a437ebb0d51296f87702e37b058c/ruleset/dockerhub/org.go#L93-L94

@bplaxco What do you think?

README.md
```diff
@@ -424,6 +454,190 @@ fragment = section of data gitleaks is looking at
...
+```
+
+##### Named Captures Example — AWS Access Key
+
+Named capture groups let a single regex extract multiple parts of a secret:
+
+```toml
+[[rules]]
+id = "aws-access-key"
+regex = '''(?P<key_id>AKIA[0-9A-Z]{16})[\s\S]{0,64}(?P<key_secret>[A-Za-z0-9/+=]{40})'''
```
rgmz: We should look for a better example: AWS is better as a composite rule.

I also don't think this would work, doesn't AWS require custom signing/hashing parameters?

README.md
```diff
@@ -424,6 +454,190 @@ fragment = section of data gitleaks is looking at
...
+
+Always include a **catch-all clause** as the last entry — a clause with no conditions. Without one, unexpected responses (e.g., 429 rate-limit, 500 server error) silently become `UNKNOWN` and can be hard to debug. Betterleaks will log a warning if the last clause is not a catch-all.
+
+**Match clause fields:**
+
+| Field | Type | Description |
+|---|---|---|
+| `status` | int or list of ints | Response status code must be one of these values. `status = 200` and `status = [200, 201]` are both valid. |
+| `words` | list of strings | Body must contain at least one of these strings (any match). Case-insensitive. |
+| `words_all` | bool | If `true`, body must contain **all** `words` |
```
rgmz, line 553: Would `words_any` be clearer?

README.md
```diff
@@ -424,6 +454,190 @@ fragment = section of data gitleaks is looking at
...
+| `negative_words` | list of strings | Body must **not** contain any of these strings. Case-insensitive. |
+| `json` | inline table | GJSON path assertions that must all be satisfied. Keys are [GJSON paths](https://github.com/tidwall/gjson/blob/master/SYNTAX.md), values can be: a scalar for exact match, `"!empty"` for existence check, or a list for one-of matching (e.g. `error = ["revoked", "inactive"]`). |
+| `headers` | inline table | Response header assertions. Keys are header names (case-insensitive), values are expected substrings (case-insensitive). |
+| `result` | string | **Required.** One of: `confirmed`, `invalid`, `revoked`, `unknown`, `error` |
```
rgmz: Are you set on `confirmed` instead of `valid`?

cmd/root.go
rgmz: > 	rootCmd.PersistentFlags().String("validation-status", "", "comma-separated list of validation statuses to include: confirmed, invalid, revoked, error, unknown")

The only concern is how we might add new output (or whatever you want to call it) flags in the future that don't directly relate to validation. I shared an example on Discord.

validate/cache.go
```diff
@@ -0,0 +1,95 @@
...
+	"sync"
+)
+
+// CachedResponse stores the status code, body, and headers of an HTTP response.
+type CachedResponse struct {
+	StatusCode int
+	Body       []byte
+	Headers    http.Header
+	Err        error
+}
```
rgmz: Could we simplify things by caching the validation result, rather than the responses? Cache rule+secret -> status + extracted attributes.

README.md
zricethezav: eh not really. `validate-full-response` feels like we are performing some action on the full response rather than just logging it.

`validate-extract-empty` could be `validation-extract-empty`

I suppose everything _could_ be `validation`?

README.md
zricethezav: > Personally, unmatched producing 'unknown' and logging a warning seems like a better UX. Manually defining a catch-all per rule can easily degrade into noise.

I can get down with that

README.md
zricethezav: You're absolutely right!

README.md
zricethezav: `words_all` and `words_any`? Replacing `words_all` with `words_any` doesnt make sense to me

README.md
zricethezav: nope! `valid` is probably better since we are using it a lot in the flags and config

README.md
rgmz: > words_all and words_any? 

Err, yeah, that. I guess I selected both options.

config/validate.go
```diff
@@ -0,0 +1,289 @@
...
+	for i, c := range v.Match {
+		if !clauseMatches(c, statusCode, body, headers) {
+			continue
+		}
+		extractMap := v.Extract
+		if len(c.Extract) > 0 {
+			extractMap = c.Extract
+		}
+		extracted := extractValues(body, headers, extractMap, includeEmpty)
+		return c.Result, extracted, fmt.Sprintf("match[%d] (%s)", i, c.Result)
```
rgmz: I don't _love_ the output being:
```
Validation:  VALID  (match[0] (valid))
```

Hopefully we can devise something more user-friendly (e.g., `status 200` or some body content).

cmd/generate/config/rules/mailgun.go
```diff
@@ -6,6 +6,22 @@ import (
 	"github.com/betterleaks/betterleaks/config"
 )
 
+func mailgunPrivateValidation() *config.Validation {
```
rgmz, line 9: Test and this works. It'd be cool if we had a way to map items from a list, in addition to plain key/value pairs (e.g., extract the list of domains returned in the response).

cmd/generate/config/rules/openai.go
```diff
@@ -6,6 +6,21 @@ import (
...
+	return &config.Validation{
+		Type:   config.ValidationTypeHTTP,
+		Method: "GET",
+		URL:    "https://api.openai.com/v1/models",
+		Headers: map[string]string{
+			"Authorization": "Bearer {{ secret }}",
+		},
+		Match: []config.MatchClause{
+			{StatusCodes: []int{200}, Result: "valid"},
+			{StatusCodes: []int{401, 403}, Result: "invalid"},
```
rgmz: We could tighten this up to be more specific:
```
{
  "error": {
    "message": "Incorrect API key provided: sk-96uqC***************************************eEaU. You can find your API key at https://platform.openai.com/account/api-keys.",
    "type": "invalid_request_error",
    "param": null,
    "code": "invalid_api_key"
  }
}
```

cmd/generate/config/rules/huggingface.go
```diff
@@ -8,6 +8,21 @@ import (
 	"github.com/betterleaks/betterleaks/config"
 )
 
+func huggingfaceValidation() *config.Validation {
```
rgmz, line 11: Huggingface has aggressive rate-limiting. Should we automatically handle 429?
```
validation:  UNKNOWN  (no match clause satisfied (status=429))
             Full response:
<!DOCTYPE html>
<html class="" lang="en">
<head>
    <meta charset="utf-8" />
    <meta
            name="viewport"
            content="width=device-width, initial-scale=1.0, user-scalable=no"
    />
    <meta
            name="description"
            content="We're on a journey to advance and democratize artificial intelligence through open source and open science."
    />
    <meta property="fb:app_id" content="1321688464574422" />
    <meta name="twitter:card" content="summary_large_image" />
    <meta name="twitter:site" content="@huggingface" />
    <meta
            property="og:title"
            content="Hugging Face - The AI community building the future."
    />
    <meta property="og:type" content="website" />

    <title>Hugging Face - The AI community building the future.</title>
    <style>
        body {
            margin: 0;
        }

        main {
            background-color: white;
            min-height: 100vh;
            padding: 7rem 1rem 8rem 1rem;
            text-align: center;
            font-family: Source Sans Pro, ui-sans-serif, system-ui, -apple-system,
            BlinkMacSystemFont, Segoe UI, Roboto, Helvetica Neue, Arial, Noto Sans,
            sans-serif, Apple Color Emoji, Segoe UI Emoji, Segoe UI Symbol,
            Noto Color Emoji;
        }

        img {
            width: 6rem;
            height: 6rem;
            margin: 0 auto 1rem;
        }

        h1 {
            font-size: 3.75rem;
            line-height: 1;
            color: rgba(31, 41, 55, 1);
            font-weight: 700;
            box-sizing: border-box;
            margin: 0 auto;
        }

        p, a {
            color: rgba(107, 114, 128, 1);
            font-size: 1.125rem;
            line-height: 1.75rem;
            max-width: 28rem;
            box-sizing: border-box;
            margin: 0 auto;
        }

        .dark main {
            background-color: rgb(11, 15, 25);
        }
        .dark h1 {
            color: rgb(209, 213, 219);
        }
        .dark p, .dark a {
            color: rgb(156, 163, 175);
        }
    </style>
    <script>
        // On page load or when changing themes, best to add inline in `head` to avoid FOUC
        const key = "_tb_global_settings";
        let theme = window.matchMedia("(prefers-color-scheme: dark)").matches
            ? "dark"
            : "light";
        try {
            const storageTheme = JSON.parse(window.localStorage.getItem(key)).theme;
            if (storageTheme) {
                theme = storageTheme === "dark" ? "dark" : "light";
            }
        } catch (e) {}
        if (theme === "dark") {
            document.documentElement.classList.add("dark");
        } else {
            document.documentElement.classList.remove("dark");
        }
    </script>
</head>

<body>
<main>
    <img
            src="https://cdn-media.huggingface.co/assets/huggingface_logo.svg"
            alt=""
    />
    <div>
        <h1>429</h1>
        <p>We had to rate limit you. Upgrade to a paid plan (https://hf.co/pricing) or send us an email at website@huggingface.co</p>
    </div>
</main>
</body>
</html>
```

cmd/generate/config/rules/huggingface.go
```diff
@@ -8,6 +8,21 @@ import (
...
+	return &config.Validation{
+		Type:   config.ValidationTypeHTTP,
+		Method: "GET",
+		URL:    "https://huggingface.co/api/whoami-v2",
+		Headers: map[string]string{
+			"Authorization": "Bearer {{ secret }}",
+		},
+		Match: []config.MatchClause{
+			{StatusCodes: []int{200}, Words: []string{`"name"`, `"id"`}, WordsAll: true, Result: "valid"},
+			{StatusCodes: []int{401, 403}, Result: "invalid"},
```
rgmz, line 22: HF also has an expired response:
```
Status:      Revoked ({"status":401,"error":"User Access Token \"datascience\" is expired"})
```

cmd/generate/config/rules/huggingface.go
zricethezav: 100%. and this brings up a good point where we should at least surface the status code and maybe a bit of the body (could just be extract).

`ERROR 429 "rate limit"` ?

README.md
zricethezav: ahh wait I think we're talking past each other here. I like `words` being the default case as OR which can be modified to AND w/  `words_all` like it is.


```golang
	Words         []string          // if set, body must contain these words (any by default)
	WordsAll      bool              // if true, ALL words must be present
	NegativeWords []string          // if set, body must NOT contain any of these
```

config/validate.go
```diff
@@ -0,0 +1,320 @@
...
+		actual := headers.Get(name)
+		if !strings.Contains(strings.ToLower(actual), strings.ToLower(expected)) {
+			return false
+		}
+	}
+
+	return true
+}
+
+// clauseReason builds a human-readable reason string from a matched clause.
```
rgmz, line 205: What if we let people define custom reasons with Liquid templates? 🤔

zricethezav: Give the people what they want

zricethezav: but considering how large this PR is, maybe let's wait on this feature

rgmz: I'm going to keep any unsolved/future stuff as unresolved, then open them as issues once this is merged.

cmd/root.go
```diff
@@ -326,6 +334,7 @@ func Detector(cmd *cobra.Command, cfg config.Config, source string) *detect.Dete
 	if detector.Verbose, err = cmd.Flags().GetBool("verbose"); err != nil {
 		logging.Fatal().Err(err).Send()
 	}
+	detector.ValidationStatusFilter, _ = cmd.Flags().GetString("validation-status")
```
rgmz, line 346: It's a bit of a code smell for this to have the raw string. `ValidationStatusFilter` should be a `map[ValidationStatus]bool` imo.

cmd/root.go
```diff
@@ -94,6 +95,13 @@ func init() {
 	rootCmd.PersistentFlags().Int("timeout", 0, "set a timeout for gitleaks commands in seconds (default \"0\", no timeout is set)")
 	rootCmd.PersistentFlags().String("regexp-engine", "re2", "regex engine (stdlib, re2)")
 
+	// Validation flags
+	rootCmd.PersistentFlags().Bool("validation", true, "enable validation of findings against live APIs")
```
rgmz, line 104: Should this be disabled by default for the initial release? I suppose it'll be gated behind an experiment flag anyway.

cmd/root.go
```diff
@@ -483,12 +503,96 @@ func bytesConvert(bytes uint64) string {
...
+	}
+	return false
+}
+
+// filterByStatus filters findings to those whose ValidationStatus matches one
+// of the comma-separated status names. Status names are case-insensitive.
+// The pseudo-status "none" matches findings from rules that have no validation
+// block (ValidationStatus == ""). Without "none", such findings are excluded
+// when any filter is active.
+func filterByStatus(findings []report.Finding, statusCSV string) []report.Finding {
```
rgmz, line 541: Related to my earlier comment about `ValidationStatusFilter`, this feels like it should be in detect.

cmd/generate/config/rules/github.go
```diff
@@ -7,6 +7,31 @@ import (
 	"github.com/betterleaks/betterleaks/regexp"
 )
 
+func githubTokenValidation() *config.Validation {
+	return &config.Validation{
+		Type:   config.ValidationTypeHTTP,
+		Method: "GET",
+		URL:    "https://api.github.com/user",
```
bplaxco, line 14: @zricethezav aside:

I wonder if the [rate_limit API](https://docs.github.com/en/rest/rate-limit/rate-limit?apiVersion=2022-11-28) would be good for this since, "Accessing [it] does not count against your REST API rate limit.", I'm pretty sure you get a 401 with bad creds, and also works with GitHub App installation access tokens.

cmd/root.go
```diff
@@ -453,6 +462,17 @@ func Detector(cmd *cobra.Command, cfg config.Config, source string) *detect.Dete
...
 
+	// Start async validation worker pool if any rules have validate blocks.
+	enableValidation := mustGetBoolFlag(cmd, "validation")
+	if enableValidation && hasAnyValidationRule(cfg) {
+		v := validate.NewValidator(cfg)
+		v.RequestTimeout, _ = cmd.Flags().GetDuration("validation-timeout")
+		v.FullResponse, _ = cmd.Flags().GetBool("validation-full-response")
+		v.ExtractEmpty, _ = cmd.Flags().GetBool("validation-extract-empty")
+		v.IncludeRequests, _ = cmd.Flags().GetBool("validation-include-requests")
+		detector.StartValidation(cmd.Context(), v, 10)
```
bplaxco, line 494: Would `validation-workers` with a default of 10 be a good flag to add as well?

cmd/root.go
bplaxco: It does feel a bit odd to have the filtering logic here. Though instead of detect why not have something in validate that you pass the findings to and then the validated and potentially filtered findings come out? Kind of more of a pipeline `fetch -> detect -> validate -> report`  and things are decoupled a bit.

config/config.go
```diff
@@ -87,6 +88,27 @@ type viperRequired struct {
 	WithinColumns *int `mapstructure:"withinColumns"`
 }
 
+type viperValidation struct {
+	Type    string             `mapstructure:"type"`
```
bplaxco, line 92: Just thought of a neat thing about type. If we need to do multi-step validation in the future, we can just extend the type to indicate it:

```toml
[rules.validate]
type = "composite" # not suggesting that term, just needed one for the example

[[rules.validate.step]]
varname = "name_for_ref_in_other_steps"
# <insert normal validation fields here, and this could also be composite if needed>
```

Also do we want to go with type or kind here? I figure it doesn't make much difference but we had used kind a few places internally and wasn't sure if we wanted to make "type" or "kind" a project standard term.

config/validate.go
```diff
@@ -0,0 +1,320 @@
...
+var validMethods = map[string]struct{}{
+	"GET":     {},
+	"POST":    {},
+	"PUT":     {},
+	"PATCH":   {},
+	"HEAD":    {},
+	"OPTIONS": {},
+}
+
+type ValidationType string
```
bplaxco, line 25: Same kind vs type question ^\_^ https://github.com/betterleaks/betterleaks/pull/24/changes#r2869999993

detect/utils.go
```diff
@@ -299,5 +302,57 @@ func printFinding(f report.Finding, noColor bool) {
 		fmt.Printf("%-12s\n%s\n", "Context:", formatMatchContext(f.MatchContext, f.Match, f.Secret, noColor))
 	}
 	f.PrintRequiredFindings()
+	printValidation(f, noColor)
 	fmt.Println("")
 }
+
+// printValidation prints the validation status block when validation has run.
+func printValidation(f report.Finding, noColor bool) {
```
bplaxco, line 310: Is there a reason to have code displaying stuff inside the detector? Would it make more sense to have display code over in the `report" or `cmd` modules?

config/validate.go
```diff
@@ -0,0 +1,320 @@
...
+}
+
+// MatchClause is one branch in a first-match-wins decision list.
+// All specified fields must be satisfied for the clause to match.
+// The first matching clause determines the finding's ValidationStatus.
+type MatchClause struct {
+	StatusCodes   []int             // if set, response status code must be one of these
+	Words         []string          // if set, body must contain these words (any by default)
+	WordsAll      bool              // if true, ALL words must be present
+	NegativeWords []string          // if set, body must NOT contain any of these
```
bplaxco, line 53: Thoughts on `NotWords`? Just sounds more boolean'y to me and is already kinda there in the field's comment.

config/config.go
```diff
@@ -575,3 +608,89 @@ func (c *Config) extend(extensionConfig Config) {
...
+	}
+	if err := v.Check(); err != nil {
+		return nil, err
+	}
+	return v, nil
+}
+
+// parseStatusCodes converts the flexible TOML status field (int, float64, or
+// list) into a []int for MatchClause.StatusCodes.
+func parseStatusCodes(raw any) ([]int, error) {
```
bplaxco, line 666: If you wanted to save a few lines of code, I wouldn't be sad typing in:

```toml
    match = [
        { statusCodes = [200], json = { active = true }, result = "valid" },
        { statusCodes = [401], result = "invalid" },
    ]
```

For some reasons fields that can be either a single value or lists always felt weird to me, but that's all just personal preference.

config/validate.go
```diff
@@ -0,0 +1,320 @@
...
+		for _, code := range c.StatusCodes {
+			if statusCode == code {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return false
+		}
+	}
```
bplaxco:
```suggestion
	if !slices.Contains(c.StatusCodes, code) {
	    return false
	}
```

config/validate.go
```diff
@@ -0,0 +1,320 @@
...
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/osteele/liquid"
+	"github.com/tidwall/gjson"
```
bplaxco, line 13: Just curious what drew you to gjson over jq, JSONPath or JMESPath? Not saying any of those are better, but just learning about gjson now ^_^

config/validate.go
```diff
@@ -0,0 +1,320 @@
+package config
```
bplaxco, line 1: Given the amount of code and the talk about future flexibility, just throwing in my 2 cents on what it could look like with my lord and savior Rego lol:


The GitHub one for example:

```toml
[[rules]]
id = "github-fine-grained-pat"
description = "Found a GitHub Fine-Grained Personal Access Token, risking unauthorized repository access and code manipulation."
regex = '''github_pat_\w{82}'''
entropy = 3
keywords = ["github_pat_"]
[rules.validate]
type = "http"
method = "GET"
url = "https://api.github.com/user"
headers = { Accept = "application/vnd.github+json", Authorization = 'token {{ secret }}' }
extract = { name = "json:name", scopes = "header:X-OAuth-Scopes", username = "json:login" }
match = [
    { status = 200, json = { id = "!empty", login = "!empty" }, result = "valid" },
    { status = [401, 403], result = "invalid" },
]
```

Could be this with just calling the rego engine on the code, and it has a compile stage that you can do to validate that the rego is good and only compiled once during initial load.

```toml
[[rules]]
id = "github-fine-grained-pat"
description = "Found a GitHub Fine-Grained Personal Access Token, risking unauthorized repository access and code manipulation."
regex = '''github_pat_\w{82}'''
entropy = 3
keywords = ["github_pat_"]
validate = '''
package betterleaks.validation

r := http.send({
      "url": "https://api.github.com/user",
      "method": "GET",
      "headers": {"Accept": "application/vnd.github+json", "Authorization": sprintf("token %s", [input.finding.secret])},
})

extract := { "name": r.body.name, "username": r.body.login, "scopes" : r.header["X-OAuth-Scopes"]}
valid if { r.status_code == 200; r.body.id; r.body.login }
invalid if [401, 403][_] == r.status
}
'''
```

And I _think_ this gets rid of the need to pull in a template engine and everything because you can just pass the context in as vars on `input`. In LeakTK we plan to just pass the whole set of findings to the Rego engine and let it do all of the verification work and spit something out at the end. So it can handle extra filtering and similar things too. The main downside is the rego policy is a different file that way but we have a thing that compiles them all into one thing that we'd download.

validate/liquid.go
```diff
@@ -0,0 +1,169 @@
...
+		decoded, err := base64.StdEncoding.DecodeString(input)
+		if err != nil {
+			return input
+		}
+		return string(decoded)
+	})
+
+	e.RegisterFilter("url_encode", func(input string) string {
+		return url.QueryEscape(input)
+	})
```
bplaxco, line 116: Would it be good to make the naming convention for the functions more consistent?

Maybe `base64_encode`, `base64_decode`?

config/config.go
rgmz: Tbh I don't really want to encourage people to specify multiple as a go-to. It's a useful feature to have, but doing it by default leads to a lot of bad/incorrect validation logic (in my observations).

cmd/root.go
zricethezav: 110% it would

config/config.go
rgmz: Heck, if it was up to me I'd make status-only matches an error.

cmd/generate/config/rules/github.go
rgmz: I don't love that, it reduces the usefulness of the information returned.

config/validate.go
rgmz: Also through `WordsNone` out there.

validate/validate.go
```diff
@@ -0,0 +1,358 @@
...
+
+	// HTTPRequests counts actual outbound HTTP requests (cache misses after singleflight).
+	HTTPRequests atomic.Int64
+}
+
+// NewValidator creates a Validator with sensible defaults.
+func NewValidator(cfg config.Config) *Validator {
+	return &Validator{
+		Config:         cfg,
+		HTTPClient:     &http.Client{},
```
rgmz, line 50: We'll need to update the client, the default settings aren't always desirable. This can be a TODO.

validate/validate.go
```diff
@@ -0,0 +1,358 @@
...
+	if err != nil {
+		if v.IncludeRequests {
+			logging.Info().Err(err).
+				Str("method", method).
+				Str("url", url).
+				Msg("validation request failed")
+		}
+		return nil, err
+	}
+	defer resp.Body.Close()
```
rgmz: I believe we'd still need this, as LimitReader could prevent the entire body from being read.
```suggestion
	defer func() { 
	  _, _ = io.ReadAll(io.Discard, res.Body)
	  _ = resp.Body.Close()
    }()
```

rgmz (commented):
Time for my daily piecemeal review.

config/validate.go
zricethezav: eh, figured we didn't need a full-blown JSONPath or jq and could reach for something native like gjson which is good at one-liners. I wanted json extraction to be fast and simple

cmd/generate/config/rules/github.go
zricethezav: I agree a/ @rgmz on this one. We cache requests too so in theory we shouldn't be hammering a single accounts `/user`

PR #26, opened 2026-02-26
Correct program name

whiteadam:
Just a small edit to change `gitleaks` to `betterleaks` in the help output.


zricethezav:
Love it, thanks!

PR #32, opened 2026-03-04
CEL for validation

zricethezav:
This PR has more deletions than additions and therefore must be good

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,30 +307,6 @@ id = "gitlab-pat"
     regexTarget = "line"
     regexes = [ '''MY-glpat-''' ]
 
-# Optional: validate whether a detected secret is live by firing an HTTP request.
-# The implicit {{ secret }} variable always contains the captured secret.
-# Named capture groups and Liquid filters are also supported.
-[[rules]]
-id = "awesome-rule-1-validated"
-description = "awesome rule 1 but validated"
-regex = '''awesome-secret-([a-zA-Z0-9]{32})'''
-keywords = ["awesome-secret-"]
-
-    [rules.validate]
-    type = "http"
-    method = "GET"
-    url = "https://api.example.com/v1/verify"
-    headers = { Authorization = "Token {{ secret }}" }
-    extract = { user = "json:user.email", scopes = "header:X-OAuth-Scopes" }
-
-    # match is a first-match-wins list; the first clause whose conditions all
-    # pass determines the finding status.
-    match = [
-        { status = 200, json = { active = true }, result = "valid" },
-        { status = 401, result = "invalid" },
-    ]
-
-
 # Global allowlists have a higher order of precedence than rule-specific allowlists.
 # If a commit listed in the `commits` field below is encountered then that commit will be skipped and no
 # secrets will be detected for said commit. The same logic applies for regexes and paths.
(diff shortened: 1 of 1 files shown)


cmd/generate/config/rules/github.go
```diff
@@ -108,7 +99,7 @@ func GitHubApp() *config.Rule {
 		Entropy:     3,
 		Keywords:    []string{"ghu_", "ghs_"},
 		Allowlists:  githubAllowlist,
-		Validation:  githubTokenValidation(),
+		ValidateCEL: githubTokenCEL,
```
rgmz: `ghs_`  requires a slightly different validation approach. I've used `/app/installations`.
https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app-installation

cmd/generate/config/rules/huggingface.go
```diff
@@ -39,7 +22,24 @@ func HuggingFaceAccessToken() *config.Rule {
...
+    "username": r.json.?name.orValue("")
+  } : r.status == 401 && r.body.contains("expired") ? {
+    "result": "revoked",
+    "reason": "Token expired"
+  } : r.status in [401, 403] ? {
+    "result": "invalid",
+    "reason": "Unauthorized"
+  } : r.status == 429 && r.body.contains("rate limit") ? {
+    "result": "error",
+    "reason": "Rate limited"
```
rgmz: Could we handle 429s as a default case in unknown?

validate/cel.go
```diff
@@ -0,0 +1,407 @@
...
+		return prg, nil
+	}
+	e.mu.RUnlock()
+
+	ast, issues := e.env.Compile(expression)
+	if issues.Err() != nil {
+		return nil, fmt.Errorf("CEL compile error:\n%s", issues.String())
+	}
+
+	prg, err := e.env.Program(ast)
```
rgmz, line 160: I recall reading that you should, if possible, run both `.Compile` and `.Check`. I don't remember why.
https://github.com/rgmz/lagotto/blob/205f5413b36333fbcf4e08f7ed998754fee070a8/core/rule/whitelist/options.go#L180-L191

cmd/generate/config/rules/huggingface.go
zricethezav: Like in the `unknown` function have a status check that looks for 429?

validate/cel.go
zricethezav:
```golang
// Compile combines the Parse and Check phases CEL program compilation to produce an Ast and
// associated issues.
//
// If an error is encountered during parsing the Compile step will not continue with the Check
// phase. If non-error issues are encountered during Parse, they may be combined with any issues
// discovered during Check.
//
// Note, for parse-only uses of CEL use Parse.
func (e *Env) Compile(txt string) (*Ast, *Issues) {
	return e.CompileSource(common.NewTextSource(txt))
}

// CompileSource combines the Parse and Check phases CEL program compilation to produce an Ast and
// associated issues.
//
// If an error is encountered during parsing the CompileSource step will not continue with the
// Check phase. If non-error issues are encountered during Parse, they may be combined with any
// issues discovered during Check.
//
// Note, for parse-only uses of CEL use Parse.
func (e *Env) CompileSource(src Source) (*Ast, *Issues) {
	ast, iss := e.ParseSource(src)
	if iss.Err() != nil {
		return nil, iss
	}
	checked, iss2 := e.Check(ast)
	if iss2.Err() != nil {
		return nil, iss2
	}
	return checked, iss2
}

```

Compile calls Check so calling both would be a waste

rgmz: My memory fails me. It's *Parse* and Check.
https://codelabs.developers.google.com/codelabs/cel-go#3

zricethezav: > Note: In this codelab, parse and check phases have been bundled into the compile() helper which can be used for the remainder of the exercises.

validate/pool.go
```diff
@@ -0,0 +1,189 @@
...
+		if !state.hasBest || newPri < statusPriority[state.best.Status] {
+			state.best = vr
+			state.hasBest = true
+		}
+
+		// Resolve when all combos done OR "valid" short-circuits.
+		shouldFire := false
+		if !state.resolved && (state.pending <= 0 || vr.Status == "valid") {
+			state.resolved = true
+			shouldFire = true
```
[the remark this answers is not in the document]
zricethezav: handled

cmd/generate/config/rules/huggingface.go
rgmz: Yeah, something small to reduce boilerplate since 429 is a well-known status code.

zricethezav: fair point fair point. Although in this case do you think having the `r.body.contains("rate limit")` is necessary for HF?

rgmz: Nah, I don't think so.

zricethezav: better name might be `defaultHTTP` instead of `unknown`?

zricethezav: Eh, `unknown` is probably fine for now. We can always change it later

PR #38, opened 2026-03-10
bugfix - printFinding output

ahrav:
This fixes the`printFinding` output when the same secret appears multiple times on one line. The code highlighted the first occurrence because it used a plain `strings.Index`, ignoring the `StartColumn` that each finding carries.

The core algorithm, if you can call it that, is simple: first try the exact position reported by the regex engine. If whitespace trimming shifted the match, scan to the right from that position. If it is not to the right, scan to the left. If there is no position hint at all, fall back to scanning from the start, as before. I know what you're thinking... super fancy :)

Opus 4.6 was used for the tests. `locateMatch` was mostly organic human grade Go.

Screenshot 2026-03-09 at 21 31 11

### DIFF
--- a/detect/utils.go
+++ b/detect/utils.go
@@ -16,6 +16,43 @@ import (
 	"github.com/charmbracelet/lipgloss"
 )
 
+// locateMatch returns the byte index of match within rawLine, using startCol
+// (1-indexed byte offset) to disambiguate duplicate occurrences. When the
+// exact position doesn't match, it searches forward then backward from the
+// expected position before falling back to the first occurrence.
+func locateMatch(rawLine, rawMatch string, startCol int) int {
+	if rawLine == "" || rawMatch == "" {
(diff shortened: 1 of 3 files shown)


detect/utils.go
```diff
@@ -210,7 +247,7 @@ func printFinding(f report.Finding, noColor bool) {
 
 	// Matches from filenames do not have a |line| or |secret|
 	if !isFileMatch {
-		matchInLineIDX := strings.Index(f.Line, f.Match)
+		matchInLineIDX := locateMatch(f.Line, f.Match, f.StartColumn)
```
ahrav, line 250: This change isn't directly related, but this boundary check clamped `lineEndIdx` when the match was near the end of the line, swallowing up to 1 trailing character of context; the new check only clamps when `lineEndIdx` actually exceeds the line length.

Issue #41, opened 2026-03-10
Split out Validate into CEL and Validate(or) packages


zricethezav:
V2 and maybe some future V1 will include CEL-based filtering at the resource (file, git commit, etc) level and rule level. All the CEL code right now lives under the `Validate` package.
```
CEL --> Validate logic (http GET, http POST, AWS db connectors, etc)
  | ___> Content/Resource filters (md5, matchAny, token efficiency, etc)
```

CEL bindings should be defined in these new packages.

PR #42, opened 2026-03-11
split out the `validate` package to `celenv` and `validate`

zricethezav:
We will be expanding the scope of CEL to handle custom filtering which means throwing all the CEL plumbing in package `validate` doesn't make sense. This PR adds a `celenv` (name tbd, could be `expression` or something similar,  but wanted to avoid `cel` b/c of `go-cel/cel`). 

package `validate` is responsible for the validate worker pool, result cache, and generating request combinations. We could rename this package to `validatorPool` to be more specific. 

package `celenv` is responsible for setting up the CEL environment (compiling and evaluating CEL programs), and CEL bindings

I also removed short circuiting on the first "valid" response in a finding "combo" group. Not only is it hard to book keep and grok the concept of "group" but it's somewhat incomplete since we would be missing potential valid findings (1 secret w/ many valid client ids, for example)

https://github.com/betterleaks/betterleaks/issues/41

### DIFF
--- a/celenv/bindings_http.go
+++ b/celenv/bindings_http.go
@@ -131,6 +131,8 @@ func (e *Environment) captureDebug(method, url, reqBody string, req *http.Reques
 // buildResponseMap constructs the CEL map returned by http.get / http.post.
 func buildResponseMap(statusCode int, body []byte, header http.Header) ref.Val {
 	var jsonBody any
+
+	// TODO log err
 	if err := json.Unmarshal(body, &jsonBody); err != nil {
 		jsonBody = map[string]any{}
 	}


detect/detect.go
```diff
@@ -948,39 +948,39 @@ func (d *Detector) AddFinding(finding report.Finding) {
...
+			if caps, ok := captureIndex[captureKey{ruleID, secret}]; ok {
+				for name, val := range caps {
+					expanded[ruleID+":"+name] = val
 				}
 			}
-			d.ValidationPool.Submit(groupID, finding, rule.CelProgram(), finding.CaptureGroups, expanded, len(combos))
 		}
-	} else {
-		d.ValidationPool.Submit(groupID, finding, rule.CelProgram(), finding.CaptureGroups, nil, 1)
+		d.ValidationPool.Submit(finding, rule.CelProgram(), finding.CaptureGroups, expanded)
```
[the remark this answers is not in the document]
zricethezav: This is fine for now. The fingerprint system is a little bunk and not designed for required rules so adding any sort of half-assed solution here isn't worth it. We need to revisit fingerprinting in v2.

celenv/bindings_http.go
```diff
@@ -0,0 +1,151 @@
...
+	for k := range resp.Header {
+		e.debugMeta["resp_header_"+strings.ToLower(k)] = resp.Header.Get(k)
+	}
+}
+
+// buildResponseMap constructs the CEL map returned by http.get / http.post.
+func buildResponseMap(statusCode int, body []byte, header http.Header) ref.Val {
+	var jsonBody any
+	if err := json.Unmarshal(body, &jsonBody); err != nil {
+		jsonBody = map[string]any{}
```
rgmz, line 137: We should still log the error, even if at a debug/trace level to avoid silently swallowing issues. We could also do a basic sanity check of validating the Content-Type header.

validate/pool.go
```diff
@@ -2,83 +2,47 @@ package validate
...
-	Meta   map[string]any
-}
-
 // validationJob is the internal unit of work for the pool.
 type validationJob struct {
-	groupID  string
-	finding  report.Finding
-	program  any // cel.Program
+	finding report.Finding
+	program any // cel.Program
```
rgmz: Why do we need the `any` type here? (I know this is an existing thing.)

zricethezav: We dont need it to be `any`, good catch

celenv/bindings_http.go
zricethezav: gonna add a TODO. We can follow up on best practices for logging in cel bindings

PR #50, opened 2026-03-16
Enrich `--match-context` (lines and columns)

zricethezav:
This PR add some nifty improvements to `--match-context` for post-scan processing of findings. 

The `--match-context` flag controls how much surrounding content is shown around
each finding in verbose (`-v`) output. It accepts a comma-separated list of
constraints using three unit types:
| Unit | Meaning |
|------|---------|
| `C`  | Columns/characters. On its own this is offset mode (characters before/after the match). Combined with `L`, it clips each line to a column window around the match position. |
| `L`  | Lines. All L values include the match line (e.g. `1L` = just the match line). Combine with `C` to build a rectangular box around the match. |

Prefix with `-` for before-only or `+` for after-only. No prefix means both
directions. A plain number with no suffix defaults to columns.

| Example | Meaning |
|---------|---------|
| `100` or `100C` | 100 columns before and after the match |
| `-128C,+16C` | 128 columns before, 16 columns after |
| `1L` | Just the line containing the match |
| `10L` | 10 lines centered on the match (match line + 9 before + 9 after) |
| `-10L,+2L` | 10 lines ending at the match, 2 lines starting at the match |
| `1L,10C` | Match line only, clipped to 10 columns before/after the match |
| `10L,500C` | 10 lines centered on the match, clipped to 500 columns before/after the match (useful as a safety cap on very long lines) |
| `-2L,+10L,-2C,+4C` | Box: 2 lines before the match, 10 lines after the match, 2 columns before match start, 4 columns after match end |

### DIFF
--- a/detect/match_context.go
+++ b/detect/match_context.go
@@ -162,8 +162,12 @@ func extractBoxContext(raw string, matchIndex []int, spec MatchContextSpec) stri
 
 	extracted := raw[ctxStart:ctxEnd]
 
-	// Box mode: apply column clipping to each line around the match column
-	if spec.ColsBefore > 0 || spec.ColsAfter > 0 {
+	// Box mode: apply column clipping to each line around the match column.
+	// Column clipping only makes sense for single-line matches; when the match
+	// spans multiple lines the first-line column offset is meaningless for
+	// subsequent lines, so we skip clipping entirely.
+	multiLine := strings.ContainsRune(raw[matchStart:matchEnd], '\n')
+	if !multiLine && (spec.ColsBefore > 0 || spec.ColsAfter > 0) {
 		matchCol := matchStart - lineStart
 		matchLen := matchEnd - matchStart
 		clipStart := max(matchCol-spec.ColsBefore, 0)
--- a/detect/match_context_test.go
+++ b/detect/match_context_test.go
@@ -157,6 +157,38 @@ L19|tttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttttt
 	}
 }
 
+func TestExtractContextMultiLineMatch(t *testing.T) {
+	// Multi-line match: column clipping should be skipped to avoid
+	// corrupting the match on subsequent lines where the column offset
+	// from the first line doesn't apply.
+	raw := "aaa\nbbbSECRET_START\nSECRET_ENDccc\nddd"
+	matchStart := strings.Index(raw, "SECRET_START")
+	matchEnd := strings.Index(raw, "SECRET_END") + len("SECRET_END")
+	matchIdx := []int{matchStart, matchEnd}
+
+	tests := []struct {
+		name string
+		spec MatchContextSpec
+		want string
+	}{
+		{
+			name: "Box: multi-line match skips col clipping",
+			spec: MatchContextSpec{Mode: ContextModeBox, LinesBefore: 1, LinesAfter: 1, ColsBefore: 2, ColsAfter: 2},
+			// All 4 lines returned unclipped because the match spans lines.
+			want: "aaa\nbbbSECRET_START\nSECRET_ENDccc\nddd",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


detect/match_context.go
```diff
@@ -0,0 +1,185 @@
...
+			ctxEnd = len(raw)
+			break
+		}
+		ctxEnd += nextNL + 1
+	}
+
+	extracted := raw[ctxStart:ctxEnd]
+
+	// Box mode: apply column clipping to each line around the match column
+	if spec.ColsBefore > 0 || spec.ColsAfter > 0 {
```
ahrav: suggestion: This breaks for multi-line matches. `clipStart`/`clipEnd` are derived from the first line’s column (`matchCol`) and then applied to all lines, but column geometry differs per line.

For a multi-line secret, the match may start at different columns (or column 0) on subsequent lines, so reusing the first line’s clip window can cut into the match itself.

Example:

```go
raw = "aaa\nbbbSECRET_START\nSECRET_ENDccc\nddd"
match = "SECRET_START\nSECRET_END"
spec = {Mode: ContextModeBox, ColsBefore: 2, ColsAfter: 2}
```

* First line: match starts at col 3 → `clipStart = 1`
* Second line: match starts at col 0, but we still apply `clipStart = 1`

Result:

```go
"SECRET_ENDccc" -> "ECRET_ENDccc" // drops 'S' from the match
```

The cor issue is treating a multi-line match as having a single column window. Clipping likely needs to be computed per line (based on that line’s own match segment), not once from the first line and reused.

3 ideas:

1. **Skip clipping for multi-line matches** (i'd probably start with this?)
   Simplest and safest. Prevents data loss, but gives up the long-line clipping cap for multi-line match context.

2. **Clip only non-match lines**
   More targeted, but adds branching and complexity.

3. **Compute clip windows per line based on the match segment on that line**
   Most correct, but also the most complex by a fair margin.


Solution 1 idea:

```go
	// Box mode: apply column clipping to each line around the match column.
	// Column clipping only makes sense for single-line matches; when the match
	// spans multiple lines the first-line column offset is meaningless for
	// subsequent lines, so we skip clipping entirely.
	multiLine := strings.IndexByte(raw[matchStart:matchEnd], '\n') >= 0
	if !multiLine && (spec.ColsBefore > 0 || spec.ColsAfter > 0) {
		matchCol := matchStart - lineStart
		matchLen := matchEnd - matchStart
		clipStart := max(matchCol-spec.ColsBefore, 0)
		clipEnd := matchCol + matchLen + spec.ColsAfter

		...
	}
```

detect/match_context.go
```diff
@@ -0,0 +1,185 @@
...
+		matchCol := matchStart - lineStart
+		matchLen := matchEnd - matchStart
+		clipStart := max(matchCol-spec.ColsBefore, 0)
+		clipEnd := matchCol + matchLen + spec.ColsAfter
+
+		lines := strings.Split(extracted, "\n")
```
ahrav, line 176: suggestion (optional): We can save a couple of allocations here by avoiding `strings.Split`/`strings.Join` and clipping in one pass with a `strings.Builder`. It’s slightly uglier, but probably a better tradeoff in this path.

Ex:
```go
var b strings.Builder
b.Grow(len(extracted))

off := 0
for off <= len(extracted) {
	nl := strings.IndexByte(extracted[off:], '\n')

	var line string
	if nl == -1 {
		line = extracted[off:]
	} else {
		line = extracted[off : off+nl]
	}

	cs := clipStart
	if len(line) <= cs {
		cs = 0 // short line: show full content rather than nothing
	}
	ce := min(clipEnd, len(line))
	b.WriteString(line[cs:ce])

	if nl == -1 {
		break
	}
	b.WriteByte('\n')
	off += nl + 1
}

extracted = b.String()
```

ahrav: I think the gain is only ~12-15%.

detect/match_context.go
ahrav: Would be good to add a test for this case as well?

zricethezav: Oh wow, good catch. Yea Let's go w/ option 1 for now

PR #53, opened 2026-03-18
AWS Validator + some things

zricethezav:
This PR introduces an AWS STS GetCallerIdentity validator using SigV4 to check if a AWS access key / secret key pair is valid. We are also moves combo/required-finding logic into the validation layer. ValidationStatus is applied to each required finding and the "best" required status bubbles up to the finding level status. 

_a poor demonstration_:
```
Finding.Validation: VALID (from required1)
 - required1.Validation: VALID
 - required2.Validation: INVALID (lower priority than required1's validation status, doesn't bubble up)
```

Had to do some tweaking w/ filters and generic required rules

### DIFF
--- a/cmd/generate/config/rules/aws.go
+++ b/cmd/generate/config/rules/aws.go
@@ -88,8 +88,7 @@ func AWSSecretAccessKey() *config.Rule {
 		SkipReport: true,
 	}
 
-	tps := utils.GenerateSampleSecrets("aws_secret_access_key", "wJalrXUtnFEMI/K7MDENG+bPxRfiCYEXAMPLEKEY")
-	tps = append(tps, utils.GenerateSampleSecrets("aws_secret_key", secrets.NewSecretWithEntropy(`[A-Za-z0-9/+=]{40}`, 4))...)
+	tps := utils.GenerateSampleSecrets("aws_secret_key", secrets.NewSecretWithEntropy(`[A-Za-z0-9/+=]{40}`, 4))
 	tps = append(tps, utils.GenerateSampleSecrets("secret_access_key", secrets.NewSecretWithEntropy(`[A-Za-z0-9/+=]{40}`, 4))...)
 	return utils.Validate(r, tps, nil)
 }


validate/result.go
```diff
@@ -47,6 +47,28 @@ func ParseResult(val ref.Val) *Result {
 	}
 }
 
+// statusPriority defines precedence for status rollup.
+// Higher value = higher priority. "valid" wins over everything; "" loses to everything.
+var statusPriority = map[string]int{
```
rgmz, line 52: If we make `Finding.ValidationStatus` an 'enum', it can be reused here for safety.

validate/pool.go
```diff
@@ -77,42 +79,130 @@ func (p *Pool) Stats() (hits, misses uint64) {
...
+				f.ValidationStatus = "error"
+				f.ValidationReason = err.Error()
+			} else {
+				f.ValidationStatus = result.Status
+				f.ValidationReason = result.Reason
+				f.ValidationMeta = result.Metadata
+			}
+			if p.FindingsCh != nil {
+				p.FindingsCh <- f
+			}
```
rgmz, line 97: Should `p.FindingsCh == nil` be an error? Or this is just a "notification" feature?

validate/pool.go
```diff
@@ -77,42 +79,130 @@ func (p *Pool) Stats() (hits, misses uint64) {
...
+				result, err = p.evalWithCacheKey(cacheKey, job.program, job.finding.Secret, merged)
+				if err != nil {
+					result = &Result{Status: "error", Reason: err.Error(), Metadata: map[string]any{}}
+				}
+				comboResults[cacheKey] = result
+			}
+
+			// Roll up finding-level status: pick the best (highest-priority) result.
+			newStatus := BetterStatus(overallStatus, result.Status)
+			if newStatus != overallStatus || bestResult == nil {
```
rgmz, line 147: Random idea: what if we gave each combo a 'score' (proximity to each other, status), and then sorted based on that? In the case where none are valid, the closest is likely the best candidate.

cmd/generate/config/rules/aws.go
```diff
@@ -48,6 +72,28 @@ func AWS() *config.Rule {
...
+			false,
+		),
+		Entropy:  4,
+		Keywords: []string{"aws_secret_access_key", "aws_secret_key", "secret_access_key"},
+		// SkipReport suppresses standalone secret-key findings; the key is
+		// always surfaced as a required component of the aws-access-token finding.
+		SkipReport: true,
+	}
+
+	tps := utils.GenerateSampleSecrets("aws_secret_access_key", "wJalrXUtnFEMI/K7MDENG+bPxRfiCYEXAMPLEKEY")
```
rgmz: This should be a FP.

cmd/generate/config/rules/aws.go
```diff
@@ -48,6 +72,28 @@ func AWS() *config.Rule {
...
+	r := config.Rule{
+		RuleID:      "aws-secret-access-key",
+		Description: "Identified an AWS secret access key, used as a component of the aws-access-token composite rule.",
+		Regex: utils.GenerateSemiGenericRegex(
+			[]string{"aws_secret_access_key", "aws_secret_key", "secret_access_key"},
+			`[A-Za-z0-9/+=]{40}`,
+			false,
+		),
+		Entropy:  4,
+		Keywords: []string{"aws_secret_access_key", "aws_secret_key", "secret_access_key"},
```
rgmz: These prefixes are too narrow to meaningfully detect most things. If you do want prefixes, we should do key/secret/token.

(Although, imo, residing in the same chunk as the access key ID + being high entropy [makes it fairly distinct](https://github.com/rgmz/lagotto/blob/70d3c7c84981a49db075d6b2059bc4c33e0e0fb2/ruleset/aws/dependencies.go#L16-L17).)

cmd/generate/config/utils/generate.go
```diff
@@ -160,3 +160,8 @@ func GenerateSampleSecrets(identifier string, secret string) []string {
 	}
 	return cases
 }
+
+// so dumb
+func IntPtr(i int) *int {
```
rgmz: Why hate generics?
```go
func Ptr[T any](v T) *T {
	return &v
}
```

zricethezav: My grugbrain didn’t think of that

validate/result.go
zricethezav: Probably should also add VALID as a global variable in the CEL environment so we arent relying on strings in CEL (🤦🏻)

zricethezav:
```
  r.status == 200 ? {
    "result": VALID,
```

vs 

```
  r.status == 200 ? {
    "result": "valid",
```

zricethezav: Let's do this in another PR

validate/pool.go
zricethezav: probably not the right place to check. 

I dont think we should even have `if p.FindingsCh != nil` now that I think it of. 

```golang
	// We have a single channel for sending findings to.
	// Findings get sent to this channel straight from
	// detectRule (non validation) OR from the ValidationPool which
	// is responsible for attempting async validation attempts.
	d.findingsCh = make(chan report.Finding, 1000)

	// little hacky. this will all be cleared up in v2
	if d.ValidationPool != nil {
		d.ValidationPool.FindingsCh = d.findingsCh
	}
```

Probably need to clean this up wire up all the channels before DetectSource is called, but that might be better for another PR. I'll remove the != nil  check in the meantime.

validate/pool.go
zricethezav: Is that different than just using the status of the most non-valid combo?

validate/result.go
```diff
@@ -47,6 +47,28 @@ func ParseResult(val ref.Val) *Result {
...
+
+// BetterStatus returns whichever of a or b has higher priority.
+// Priority order: valid > revoked > unknown > invalid > error > "".
+// This is used for rolling up per-component validation results into an
+// overall finding-level status for composite rules.
+func BetterStatus(a, b string) string {
+	if statusPriority[b] > statusPriority[a] {
+		return b
+	}
+	return a
```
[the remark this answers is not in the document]
zricethezav: Will handle in another issue

PR #59, opened 2026-03-21
feat: support multiple paths in dir command

### LINKED ISSUE #56: Support individual files to be passed for the secrets scan
It would be great to add support for passing a list of files to scan (--files?), especially when working with very large repositories. This would interact well with pre-commit hooks such as https://prek.j178.dev/.

mvanhorn:
## Summary

Allow the `dir` command (and its `file`/`directory` aliases) to accept multiple paths as arguments, scanning each independently and aggregating findings.

## Why this matters

When working with large repositories and pre-commit hooks, scanning only the staged files is much faster than scanning an entire directory. Currently users must invoke betterleaks once per file or use workarounds. This change lets users pass multiple paths in a single invocation:

```
betterleaks dir file1.py file2.js src/config/
```

## Changes

In `cmd/directory.go`:
- Changed `Use` string from `"dir [flags] [path]"` to `"dir [flags] [path...]"` to indicate variadic args
- Modified `runDirectory` to iterate over all args (defaulting to `"."` when none provided)
- Each path is scanned via `detector.DetectSource` and findings are aggregated into `allFindings`
- Errors are logged per-source with the source path for traceability

## Testing

- `go build ./...` passes
- Backward compatible: `betterleaks dir .` still works (single arg)
- New: `betterleaks dir file1.go file2.go` scans both files

Fixes #56

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/cmd/directory.go
+++ b/cmd/directory.go
@@ -52,6 +52,8 @@ func runDirectory(cmd *cobra.Command, args []string) {
 		lastDetector *detect.Detector
 		scanErr      error
 	)
+
+	totalBytes := uint64(0)
 	for _, source := range sourcesList {
 		initConfig(source)
 		cfg := Config(cmd)
@@ -75,7 +77,10 @@ func runDirectory(cmd *cobra.Command, args []string) {
 			scanErr = detectErr
 		}
 		allFindings = append(allFindings, findings...)
+		totalBytes += detector.TotalBytes.Load()
 	}
+
+	lastDetector.TotalBytes.Swap(totalBytes)
 
 	findingSummaryAndExit(lastDetector, allFindings, exitCode, start, scanErr)
 }


cmd/directory.go
```diff
@@ -15,23 +16,20 @@ func init() {
...
 func runDirectory(cmd *cobra.Command, args []string) {
-	// grab source
-	source := "."
-	if len(args) == 1 {
-		source = args[0]
-		if source == "" {
-			source = "."
-		}
+	// grab sources
+	sources_list := args
```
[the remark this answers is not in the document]
zricethezav: I agree w/ this nit, please use camelCase here

cmd/directory.go
```diff
@@ -15,23 +16,20 @@ func init() {
...
+	if len(sources_list) == 0 {
+		sources_list = []string{"."}
 	}
 
-	initConfig(source)
+	initConfig(sources_list[0])
```
zricethezav: An alternative path to what the greptile review suggests is to just create a new detector for each source. This would likely be the easier option since you dont have to manually AddGitleaksIgnore. The greptile review also misses the fact that configs can be dynamically loaded if a `.gitleaks/.betterleak.toml` file exists at the root of a directory source which is why doing a big loop w/ `initConfig` and `Detector()` is probably the easiest option.

bplaxco: +1 on spinning up a detector + config per source

rgmz: Do you think we should account for cases where you pass in `root root/subfolder` and suppress duplicates?

mvanhorn: Good point - with the per-detector-per-source approach, overlapping paths like `root root/subfolder` would produce duplicate findings since `root` already scans `root/subfolder`. Two options: deduplicate by file path + line number after aggregating results, or detect path containment upfront and skip nested paths. Both are straightforward. Happy to add either in this PR or as a follow-up - what do you prefer?

rgmz: It's not a blocking concern, I'm mostly thinking about the user experience which can be improved in the future. What do you think, @zricethezav @bplaxco ?

zricethezav: @mvanhorn sorry for missing this conversation. I think I like this option the best:

 > detect path containment upfront and skip nested paths
 
 If you're up for it, would you mind taking a crack at implementing this?

mvanhorn: Implemented in dbb50af - `removeNestedPaths` resolves each argument to its absolute path and filters out any that are children of another path in the list. So `betterleaks dir root root/subfolder` now scans only `root`.

zricethezav (approved):
nice work

mvanhorn:
Appreciate the merge - multi-path support was a natural extension.

PR #82, opened 2026-04-02
Source Attributes

zricethezav:
Source attributes paves the way for us to introduce new sources in betterleaks. The problem we had with the existing implementation was that source metadata were hardcoded fields on the fragment. That's not very flexible and would make developing new sources a chore.

This change allows us to start kicking the tires on introducing CEL filtering to replace allowlists. Stuff like:

```toml
filter = '''
  attr("git.author_email") == "ci-bot@example.com" ||
  attr("resource_kind") == "s3" && attr("s3.bucket").matches(".*-test")
'''

```

(`attr` in this example would be some custom CEL binding)


sources/fragment.go
```diff
@@ -1,28 +1,52 @@
...
 
-	// CommitInfo captures additional information about the git commit if applicable
-	CommitInfo *CommitInfo
+	// Attributes holds all source-specific metadata as flat key-value pairs.
+	Attributes []Attribute
+}
+
+// Attr returns the value for the given key, or "" if not present.
+func (f *Fragment) Attr(key string) string {
+	for _, a := range f.Attributes {
```
rgmz: I'm guessing that slices are faster / more efficient than maps? What about duplicate key/value pairs, is that something we should guard against?

sources/attribute.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+// Well-known attribute keys (constants prevent typos at call sites).
+const (
+	// universal attributes
+	AttrPath = "path"
+
+	// Git attributes
+	AttrGitSHA         = "git.sha"
+	AttrGitAuthorName  = "git.author_name"
```
rgmz, line 10: Should we include committer name and email?

sources/fragment.go
zricethezav: Yep that should absolutely be protected against. Good catch

sources/fragment.go
```diff
@@ -1,28 +1,46 @@
...
 	// StartLine is the line number this fragment starts on
 	StartLine int
 
-	// CommitInfo captures additional information about the git commit if applicable
-	CommitInfo *CommitInfo
+	// Attributes holds all source-specific metadata as flat
+	Attributes map[string]string
+}
+
+func (f *Fragment) SetAttr(key, value string) {
```
rgmz, line 25: Some questions:
1. The currently defined attribute keys are all lowercase. Should we enforce that?
2. Should we enforce some sort of format/valid characters (`[a-zA-Z0-9._-]`) or min/max length for keys?
3. Should we discard empty/blank values?

zricethezav: I think 1 and 2 would be better suited for CI rather than runtime. 

Number 3 is a toss up for me. I could see a world where people want the empty string set for a value.

sources/attribute.go
zricethezav: let's save that until we refactor and enhance the git source

PR #92, opened 2026-04-13
Dynamic attributes (prep work for more sources) and `Run()` entry (better api)

zricethezav:
This PR does a few things:

**1. Replaces `DetectSource` with `Run` as the primary entrypoint.**

- Rather than creating a new package like `pipeline` that would aim to supplant `detect`, I chose to just clean up `detect/detect.go`, move a bunch of the old "detect" entry points into a deprecated file. Since we didn't change the core detection logic this seemed like the most minimally invasive change we could do.
- `Run` replaces "DetectSource" as the main entry point. It offers basically the same api as DetectSource:
```golang
// OLD
func (d *Detector) DetectSource(ctx context.Context, source sources.Source) ([]report.Finding, error) {

// NEW
type Result struct {
	Finding report.Finding
	Err     error
}
func (d *Detector) Run(ctx context.Context, source sources.Source) iter.Seq[Result] {
```
Internally, Run uses a buffered channel between the scanning goroutine and the iterator consumer that performs the `yield` calls. This lets the caller receive findings as they are found, rather than after the entire scan finishes. Nifty, right?

**2. `Finding` now uses a map-based attribute system for reporting and filtering (currently with allowlists).**

A goal of Betterleaks is to introduce new sources like scanning GitHub PR comments and descriptions (like this one) or S3 buckets. We also want to let users filter on various source or resource level attributes (like comment author, or date, or email) which are set in the source implementation. 

Ex: setting git source attribute

```golang
commitAttrs[AttrGitMessage] = gitdiffFile.PatchHeader.Message()
```

In a later release we'll introduce ways to filter on any of the attributes set in the source. This is just the groundwork for that. Moving from a static metadata model to a dynamic one using a simple `map[string]string` helps decouple detection logic from individual source implementations and makes the finding model more flexible going forward.

P.S. sorry for the big PR.


cmd/directory.go
```diff
@@ -54,28 +51,36 @@ func runDirectory(cmd *cobra.Command, args []string) {
...
+			Path:            source,
+			Sema:            detector.Sema,
+			MaxArchiveDepth: maxArchiveDepth,
+		}
+
+		var findings []report.Finding
+		for result := range detector.Run(cmd.Context(), s) {
+			if result.Err != nil {
+				logging.Error().Err(result.Err).Msg("error scanning source")
+				continue
```
rgmz, line 75: Are there any errors we should consider as fatal? i.e., return instead of continue

sources/source.go
```diff
@@ -13,4 +13,5 @@ type Source interface {
 	// Fragments provides a filepath.WalkDir like interface for scanning the
 	// fragments in the source
 	Fragments(ctx context.Context, yield FragmentsFunc) error
+	// FindingAttrs() []string
```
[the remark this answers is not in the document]
rgmz: 👀

detect/utils.go
```diff
@@ -438,3 +439,74 @@ func sortedMapKeys(m map[string]any) []string {
...
+var lowercaseBufPool = sync.Pool{
+	New: func() any {
+		buf := make([]byte, 0, 128*1024)
+		return &buf
+	},
+}
+
+// getLowerBuf returns an ASCII-lowercased copy of s in a pooled byte buffer.
+// Caller must call putLowerBuf when done with the returned slice.
+func getLowerBuf(s string) (*[]byte, []byte) {
```
rgmz, line 454: This is sufficiently complicated and "load-bearing" that it deserves a test, imo.

detect/detect.go
```diff
@@ -247,6 +200,170 @@ func NewDetectorDefaultConfig() (*Detector, error) {
...
+
+			if !yield(res) {
+				cancel()
+				return
+			}
+		}
+	}
+}
+
+func (d *Detector) routeFinding(finding report.Finding, emit func(Result) error) error {
```
rgmz: Can you add a short comment to this? 'Route' isn't immediately obvious to me.

report/finding.go
```diff
@@ -21,40 +23,27 @@ type Finding struct {
...
 
-	// unique identifier
-	Fingerprint string
+	// Fragment used for multi-part rule checking and CEL filtering
+	Fragment *sources.Fragment `json:",omitempty"`
 
-	// MatchContext contains surrounding lines around the match
-	MatchContext string `json:",omitempty"`
+	// Metadata
+	Attributes map[string]string `json:",omitempty"`
```
rgmz, line 47: What's the difference between `Attributes` and `ValidationMeta`? Are attributes strictly provenance?

report/finding.go
```diff
@@ -272,3 +288,308 @@ func formatSetStatus(status string, noColor bool) string {
...
+		if f.MatchContext != "" {
+			fmt.Printf("%-12s\n%s\n", "Context:", formatMatchContext(f.MatchContext, f.Match, f.Secret, noColor))
+		}
+		printValidation(*f, noColor)
+		f.PrintRequiredFindings(noColor, redact)
+		fmt.Println("")
+		return
+	}
+	if len(f.Tags) > 0 {
+		fmt.Printf("%-12s %s\n", "Tags:", f.Tags)
```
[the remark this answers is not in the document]
rgmz: Seems like a legitimate concern.

rgmz (commented):
Gave this a cursory review. No major concerns

report/finding.go
zricethezav: That's right. It may be worth shoving everything in `Attributes` at some point.

detect/detect.go
zricethezav: It's a bad function. The idea was the finding could be routed different ways; 1. ignored, 2. to a validator pool, 3. to the regular finding chan. Although now I'm kinda questioning the error handling here. I imagine we dont want to return an error on validation errors?
```diff
                                findings := d.detectFragment(runCtx, fragment)
                                for _, finding := range findings {
-                                       // if err := routeFinding(finding); err != nil {
-                                       if err := d.routeFinding(finding, emit); err != nil {
-                                               return err
+                                       if d.ignore(finding) {
+                                               continue
+                                       }
+                                       if d.ValidationPool != nil {
+                                               if rule, ok := d.Config.Rules[finding.RuleID]; ok && rule.CelProgram() != nil {
+                                                       if err := d.ValidationPool.SubmitContext(ctx,
+                                                               finding,
+                                                               rule.CelProgram(),
+                                                               finding.CaptureGroups); err != nil {
+
+                                                               if errors.Is(err, context.Canceled) {
+                                                                       return errStopIteration
+                                                               }
+                                                               return err
+                                                       }
+                                                       // return nil
+                                                       continue
+                                               }
                                        }
+                                       emit(Result{Finding: finding})
                                }

                                return nil
@@ -311,6 +334,9 @@ func (d *Detector) Run(ctx context.Context, source sources.Source) iter.Seq[Resu
                                if res.Finding.ValidationStatus != "" {
                                        d.ValidationCounts[res.Finding.ValidationStatus]++
                                }
+                               if !d.SkipFindingAppend {
+                                       d.findings = append(d.findings, res.Finding)
+                               }
                        }
```

cmd/directory.go
zricethezav: I can't think of any. If they pop up we can address them in another PR.

zricethezav:
Gonna give this a merge. Follow up items include:
- context handling between sources/detect/cmd
- starting on the cel filter

PR #100, opened 2026-04-20
Feat/cel filters

zricethezav:
This PR enables CEL filtering on source attributes and finding fields for ultimate configurability. We are moving away from `[[allowlists]]` and moving towards a much more expressive `filter` and `prefilter` config paradigm.

### Betterleaks `filter`s
Filters replace legacy allowlists, entropy checks, and token efficiency checks with dynamic Common Expression Language (CEL) statements. If a filter expression evaluates to `true`, the item is **skipped/discarded**.

* **`prefilter`**: Exists only at the global level. It evaluates *before* any regex runs and only has access to file/commit metadata (`attributes`). Use this to entirely bypass binary files or bot commits.
* **`filter`**: Exists globally and per-rule. It evaluates *after* a regex match is found and has access to both `attributes` and the `finding` itself.

Note that safe attribute access requires somewhat cumbersome syntax, `attributes.[?"key"].orValue("")`. If `key` does not exist in the attributes map, then it will default to using an empty string, `""`.

### Available `filter` bindings

| Binding / Function | Description |
| :--- | :--- |
| `attributes` | A map of metadata. Keys include: `path`, `git.sha`, `git.author_name`, `git.author_email`, `git.date`, `git.message`, `git.remote_url`, `git.platform`, `fs.symlink`. Full list of available keys [available here](https://github.com/betterleaks/betterleaks/blob/main/sources/attribute.go). |
| `finding` | A map representing the secret. Keys include: `secret` (the extracted value), `match` (the full regex match), `line` (the line of code), `rule_id`, and `description`. |
| `matchesAny(string, list)` | Returns `true` if the string matches any of the provided regex patterns. |
| `containsAny(string, list)` | Returns `true` if the string contains any of the provided strings (uses an efficient Aho-Corasick substring match). |
| `entropy(string)` | Returns the Shannon entropy (float) of the string. Useful for filtering out non-random placeholders. |
| `failsTokenEfficiency(string)`| Returns `true` if the string tokenizes too efficiently (i.e., it looks like natural language instead of a random secret). |
---

Example `filter` CEL expression:

```toml
filter = '''
(
    // Ignore if authored by a bot AND inside the fixtures folder AND the secret contains a known test string.
    attributes[?"git.author_name"].orValue("").endsWith("[bot]") &&
    attributes[?"path"].orValue("").startsWith("tests/fixtures/") &&
    containsAny(finding["secret"], ["_MOCK_", "_TEST_"])
)
||
(
    // Ignore if it's a Markdown or text file AND the specific line of code contains instructional text.
    matchesAny(attributes[?"path"].orValue(""), [r"""(?i)\.(?:md|txt|csv)$"""]) &&
    (
        containsAny(finding["line"], ["Example:", "Placeholder:", "Replace this with"]) ||
        finding["secret"] == "SUPER_SECRET_EXAMPLE_KEY_12345"
    )
)
||
(
    // Ignore if the entropy is low AND it tokenizes like natural language instead of a random string.
    entropy(finding["secret"]) <= 2.5 &&
    failsTokenEfficiency(finding["secret"])
)
'''
```

Existing allowlists will still work! Internally there is a translation layer that converts allowlists to equivalent CEL expressions.

---

Other random shit in this PR:
- Slimmed down the README. Eventually we'll move all the docs in `docs` to a `betterleaks.com/docs` page.
- fixed a bug where fingerprints weren't being sent to stdout


cmd/detect.go
```diff
@@ -75,7 +75,7 @@ func runDetect(cmd *cobra.Command, args []string) {
 	if noGit {
 		findings, err = detector.DetectSource(
 			cmd.Context(), &sources.Files{
-				Config:          &cfg,
+				Skip:            detector.SkipFunc(),
```
zricethezav: Might be a bit premature but this feels better than passing the config around more than it needs to be. Albeit, it's technically a breaking API change...

detect/detect.go
```diff
@@ -168,24 +184,106 @@ type Detector struct {
...
-		ValidationCounts: make(map[string]int),
-		Config:           cfg,
-		prefilter:        *ahocorasick.NewTrieBuilder().AddStrings(maps.Keys(cfg.Keywords)).Build(),
-		Sema:             semgroup.NewGroup(ctx, 40),
-		tokenizer:        tke,
+	// Compile CEL filter programs so they are available at scan time.
+	// This is the single compilation owner — callers should NOT compile separately.
+	if compileErr := cfg.CompileCELFilters(tke); compileErr != nil {
+		logging.Fatal().Err(compileErr).Msg("failed to compile CEL filters")
+	}
```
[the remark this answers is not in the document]
zricethezav: fair

detect/detect.go
```diff
@@ -348,29 +447,12 @@ func (d *Detector) Run(ctx context.Context, source sources.Source) iter.Seq[Resu
...
 			Msg("skipping finding: global fingerprint")
 		return true
-	} else if commit != "" {
-		if _, ok := d.gitleaksIgnore[finding.Fingerprint]; ok {
-			logger.Debug().
-				Str("fingerprint", finding.Fingerprint).
-				Msgf("skipping finding: fingerprint")
-			return true
-		}
 	}
```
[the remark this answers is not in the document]
zricethezav:
```go
func (f *Finding) SetFingerprint() {
	path := f.Attributes[sources.AttrPath]
	commit := f.Attributes[sources.AttrGitSHA]

	globalFingerprint := fmt.Sprintf("%s:%s:%d", path, f.RuleID, f.StartLine)
	if commit != "" {
		f.Fingerprint = fmt.Sprintf("%s:%s:%s:%d", commit, path, f.RuleID, f.StartLine)
	} else {
		f.Fingerprint = globalFingerprint
	}
}

```
bad bot

sources/git.go
```diff
@@ -390,6 +382,24 @@ func (s *Git) Fragments(ctx context.Context, yield FragmentsFunc) error {
 					commitAttrs[AttrGitAuthorName] = gitdiffFile.PatchHeader.Author.Name
 					commitAttrs[AttrGitAuthorEmail] = gitdiffFile.PatchHeader.Author.Email
 				}
+
+				// CEL prefilter (also covers commit SHAs via attributes["git.sha"]).
+				// Build a temporary map that includes the path for the prefilter check only —
+				// do NOT put the path in commitAttrs itself because archive enrichment code
```
rgmz: Could we enforce this with an explicit check + error?

config/betterleaks.toml
```diff
@@ -76,8 +78,10 @@ stopwords = [
 id = "1password-secret-key"
 description = "Uncovered a possible 1Password secret key, potentially compromising access to secrets in vaults."
 regex = '''\bA3-[A-Z0-9]{6}-(?:(?:[A-Z0-9]{11})|(?:[A-Z0-9]{6}-[A-Z0-9]{5}))-[A-Z0-9]{5}-[A-Z0-9]{5}-[A-Z0-9]{5}\b'''
-entropy = 3.8
 keywords = ["a3-"]
+filter = '''
+(entropy(finding["secret"]) <= 3.8)
```
rgmz: Does this need to be enclosed in parenthesis? It seems like unnecessary visual noise.

cmd/generate/config/rules/config.tmpl
```diff
@@ -19,23 +19,11 @@ title = "{{.Title}}"
 minVersion = "v8.25.0"
 betterleaksMinVersion = "v1.0.0"
```
rgmz: We'd need to bump this, aye?

cmd/git.go
```diff
@@ -109,11 +112,21 @@ func runGit(cmd *cobra.Command, args []string) {
 		}
 	}
 
-	findings, err = detector.DetectSource(cmd.Context(), src)
+	detector.SkipFindingAppend = true
+	for result := range detector.Run(cmd.Context(), src) {
+		if result.Err != nil {
+			if err == nil {
+				err = result.Err
```
rgmz: Do we only want to take the first error? Why not accumulate all?

config/config.go
```diff
@@ -116,7 +146,15 @@ type Extend struct {
...
 }
 
-func (vc *ViperConfig) Translate() (Config, error) {
+func (vc *ViperConfig) Translate() (*Config, error) {
+	// Top-level Translate calls must reset the global extendDepth so that
+	// translateLegacyFilters (which only runs at depth 0) isn't skipped
+	// due to leftover state from a previous Translate call (e.g. in tests).
+	isTopLevel := extendDepth == 0
+	if isTopLevel {
+		defer func() { extendDepth = 0 }()
```
rgmz: I wonder if we should guard this with a mutex. Relying on mutating  a global variable is rough. 😅

config/translate_filters.go
```diff
@@ -0,0 +1,252 @@
...
+// Called once, at the end of Translate(), after all extends and targeted allowlist
+// population are complete (extendDepth == 0). Logs translated expressions at debug level
+// so users can copy them and migrate away from the deprecated fields.
+//
+// Semantics: all CEL filter/prefilter expressions return true = skip, false = keep.
+// Prefilter exists only at the global level (attributes-only, runs before regex).
+// Rules have only filter (attributes + finding, runs per match).
+//
+// TranslateLegacyFilters converts deprecated Allowlists, Entropy, and TokenEfficiency
+// fields into CEL prefilter/filter expressions. Exported for use by the config generator.
```
rgmz: Do we need both?
```suggestion
// translateLegacyFilters converts deprecated Allowlists, Entropy, and TokenEfficiency
// fields into CEL prefilter/filter string expressions on the Config and its Rules.
...
// TranslateLegacyFilters converts deprecated Allowlists, Entropy, and TokenEfficiency
// fields into CEL prefilter/filter expressions. Exported for use by the config generator.
```

rgmz (commented):
Stream of consciousness

cmd/git.go
zricethezav: eh this one is really just to satisfy passing _something_ to `findingSummaryAndExit` which is outdated. We probably need to think through this a little more.

config/config.go
zricethezav: https://github.com/betterleaks/betterleaks/pull/100/changes/cde73916ddad0710f3e59179c43852c0cfc37cee should handle it

celenv/bindings_filter.go
```diff
@@ -0,0 +1,252 @@
...
+	for i, p := range patterns {
+		parts[i] = "(?:" + p + ")"
+	}
+	re, err := blregexp.Compile(strings.Join(parts, "|"))
+	if err != nil {
+		return nil
+	}
+	regexCache.Store(key, re)
+	return re
+}
```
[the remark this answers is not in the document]
zricethezav: This doesn't silently fail, it gets caught during the CEL eval. Sure, it'd be nice to catch it sooner. Something we could follow up with is a `config-doctor` command or `config` command which validates all aspects of the config.

config/betterleaks.toml
zricethezav: fixt

celenv/bindings_filter.go
```diff
@@ -0,0 +1,253 @@
...
+				}
+				terms := celListToStrings(rhs)
+				if len(terms) == 0 {
+					return types.Bool(false)
+				}
+				trie := getOrBuildTrie(terms)
+				if trie == nil {
+					return types.Bool(false)
+				}
+				return types.Bool(trie.MatchFirstString(strings.ToLower(string(str))) != nil)
```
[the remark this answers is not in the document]
zricethezav: ahocorasick prefilter is case insensitive, not an issue

README.md
```diff
@@ -66,504 +44,93 @@ cd betterleaks
...
-#    configuration and the default configuration the extended rules or
-#    attributes of them will override the default rules.
-#    Another thing to know with extending configurations is you can chain
-#    together multiple configuration files to a depth of 2. Allowlist arrays are
-#    appended and can contain duplicates.
+# Prefilters run BEFORE any regex matching occurs, hence the _pre_.
+# They only have access to the `attributes` map (like file path and git metadata).
+# Use this to quickly bypass heavy binary files, vendor directories, or noisy bots.
+prefilter = '''
+(matchesAny(attributes[?"path"].orValue(""), [
```
rgmz, line 62: Do we have a link to all the available functions and fields?

README.md
```diff
@@ -66,504 +44,93 @@ cd betterleaks
...
-      "result": "valid",
-      "username": r.json.?login.orValue(""),
-      "name": r.json.?name.orValue(""),
-      "scopes": r.headers[?"x-oauth-scopes"].orValue("")
-    } : r.status in [401, 403] ? {
-      "result": "invalid",
-      "reason": "Unauthorized"
-    } : unknown(r)
-  )
+cel.bind(r,
```
rgmz, line 99: Would checksum logic also go here?

rgmz (commented):
Code aside, we should support this change with docs + an AGENTS.md to aide people in writing new rules.

sources/file.go
```diff
@@ -72,9 +71,9 @@ type File struct {
 	Symlink string
 	// Buffer is used for reading the content in chunks
 	Buffer []byte
-	// Config is the gitleaks config used for shouldSkipPath. If not set, then
-	// shouldSkipPath is ignored
-	Config *config.Config
+	// Skip is a callback that decides whether to skip a file based on its
+	// attributes (e.g. path). If nil, no skipping is performed.
+	Skip SkipFunc
```
bplaxco: nitpick: Might call it something like `ShouldSkip` to avoid potential confusion on whether you're asking if it should skip or you're marking it as something to skip.

cmd/generate/config/utils/validate.go
```diff
@@ -119,5 +119,14 @@ func createSingleRuleDetector(r *config.Rule) *detect.Detector {
 			logging.Fatal().Err(err).Msg("invalid global allowlist")
 		}
 	}
-	return detect.NewDetector(cfg)
+
+	// Translate legacy allowlists/entropy/tokenEfficiency into CEL expressions
+	// and compile them so filtering works during validation.
+	if err := cfg.TranslateLegacyFilters(); err != nil {
+		logging.Fatal().Err(err).Msg("failed to translate legacy filters")
```
bplaxco: nitpick: might tweak the wording a bit to help users bridge the gap from allowlist to current filter.


```suggestion
		logging.Fatal().Err(err).Msg("failed to translate legacy allowlists to filters")
```

cmd/generate/config/utils/validate.go
```diff
@@ -119,5 +119,14 @@ func createSingleRuleDetector(r *config.Rule) *detect.Detector {
...
+	}
+	// NewDetector calls NewDetectorContext which compiles CEL filters,
+	// so we don't need a separate compilation step.
+	d := detect.NewDetector(cfg)
+	return d
```
bplaxco: style nitpick if you care: 


```suggestion
	return detect.NewDetector(cfg)
```

cmd/generate/config/main.go
```diff
@@ -423,6 +418,21 @@ func main() {
 		slices.Sort(allowlist.Commits)
 		slices.Sort(allowlist.StopWords)
 	}
+
+	// Translate legacy allowlists, entropy, and tokenEfficiency into CEL
+	// prefilter/filter expressions so the generated TOML uses native CEL.
+	if err := cfg.TranslateLegacyFilters(); err != nil {
+		logging.Fatal().Err(err).Msg("Failed to translate legacy filters to CEL")
```
bplaxco: nitpick similar to https://github.com/betterleaks/betterleaks/pull/100/changes#r3143881428 to make the errors consistent:


```suggestion
		logging.Fatal().Err(err).Msg("failed to translate legacy allowlists to filters")
```

cmd/generate/config/main.go
```diff
@@ -423,6 +418,21 @@ func main() {
...
+	}
+	// Clear deprecated fields now that they've been translated.
```
bplaxco: Would it make more sense to do these in `TranslateLegacyFilters`?

Reasons:

- Don't have to keep up with if it's done
- Downstream lib consumers can just call the function and not need to worry about editing fields

celenv/prefilter.go
```diff
@@ -0,0 +1,84 @@
...
+// PrefilterEnv is a CEL environment for evaluating prefilter expressions.
+// Only `attributes map<string,string>` is in scope.
+type PrefilterEnv struct {
+	env   *cel.Env
+	mu    sync.RWMutex
+	cache map[string]cel.Program
+}
+
+// NewPrefilterEnv creates a CEL environment for evaluating prefilter expressions.
+// TODO decide on the name Prefilter or FragmentFilter or ResourceFilter?
```
bplaxco, line 23: I think prefilter (i.e. filter before looking any file content) and filter (filter based on the findings) maps pretty clear conceptually. 👍

cmd/detect.go
bplaxco: Looking forward to v2 where that pipeline architecture gives you a clean way to pre-filter outside of the resources 👌

cmd/git.go
```diff
@@ -109,11 +112,21 @@ func runGit(cmd *cobra.Command, args []string) {
...
+			logging.Error().Err(result.Err).Msg("failed to scan Git repository")
+			continue
+		}
 
-	if err != nil {
-		// don't exit on error, just log it
-		logging.Error().Err(err).Msg("failed to scan Git repository")
+		findings = append(findings, result.Finding)
+		if verbose {
+			result.Finding.Print(noColor, redact)
```
bplaxco, line 138: That feels like it really should go in a reporting module. Feels like a bit of a code smell to put display logic in something that feels like a data struct.

cmd/git.go
bplaxco: What about something like:


Borrowing from go's lib: https://cs.opensource.google/go/go/+/refs/tags/go1.26.2:src/errors/join.go;l=20  (but not doing the newlines in the error output :P)

```go
struct multipleErrors {
    msg string
    []errs errors
}

func (e *multipleErrors) Error() string {
    return e.msg
}

func (e *multipleErrors) Unwrap() []error {
    return e.errs
}
```
```go
var detectErrors []error
//...
    if result.Err != nil {
        detectErrors = append(detectErrors, result.Err)
        logging.Error().Err(result.Err).Msg("failed to scan Git repository")
    }
// ...

if n := len(detectErrors); n > 0 {
    err = &multipleErrors {
        msg: fmt.Sprintf("%d errors encountered during detector run", n),
        errs: detectErrors,
    }
}

// ...
```

That should give you a pretty error msg, decent logs, and lib users the ability to still access each underlying error by unwrapping it using `errors.Unwrap(err)` 👍 


I would say use `errors.Join` but I don't like that it adds newlines to the output :P

config/config.go
```diff
@@ -199,6 +229,7 @@ func (vc *ViperConfig) Translate() (Config, error) {
 		}
 
 		cr.ValidateCEL = vr.Validate
+		cr.Filter = vr.Filter
```
bplaxco, line 232: Should `cr.ValidateCEL` also just be `cr.Validate` to keep things consistent between the two fields?

config/rule.go
```diff
@@ -65,6 +65,14 @@ type Rule struct {
 
 	// celProgram is the compiled CEL program, set at config load time.
 	celProgram cel.Program
```
bplaxco, line 67: For consistency:

```suggestion
	validateProgram cel.Program
```

config/rule.go
bplaxco, line 64: For consistency:


```suggestion
	Validate string
```

config/rule.go
bplaxco: Probably also want to update the other generic CEL funcs on the rule too.

detect/detect.go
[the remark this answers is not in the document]
bplaxco: Confirming, so we'll no longer support global ignores for git repos? (I'm cool with that because I prefer using the config for ignores, but that does seem like a change in behavior).

detect/detect.go
```diff
@@ -538,14 +613,8 @@ func (d *Detector) detectFragmentWithRule(fragment sources.Fragment,
 		return findings
 	}
 
-	// check if commit or file is allowed for this rule.
-	// TODO replace source-specific attribute checking with a CEL-based filter
-	if isAllowed, event := checkCommitOrPathAllowed(logger, fragment, r.Allowlists); isAllowed {
```
bplaxco, line 543: @zricethezav 

I think we should add `Prefilter` at the rule level too. I have rules where I skip whole directories because they're just not high enough fidelity to run on the files there.

The rules already have a few prefilters (e.g. ahocorasick, path, etc). What are your thoughts on taking everything that isn't the ahocorasick check and regex on a rule and have translate turn that into either `Prefilter` or `Filter`?

It could look something like this in the code and then you wouldn't have to check later about the path only rules not matching because the prefilter would have already compiled that logic.

```go
			// Use Aho-Corasick to find keyword matches, then map directly
			// to the rules that need checking via KeywordToRules.
			// Use a pooled byte buffer for lowercasing to avoid allocating
			lowerBufPtr, lowerBuf := getLowerBuf(currentRaw)
			acMatches := d.prefilter.Match(lowerBuf)

			// Build a set of rule IDs to check based on keyword matches.
			rulesToCheck := make(map[string]struct{}, len(acMatches))
			for _, m := range acMatches {
				// m.Match() returns the keyword as []byte; convert to string
				// for the map lookup. This is a small allocation per keyword
				// match (typically few), not per fragment byte.
				keyword := string(m.Match())
				for rule, ruleID := range d.Config.KeywordToRules[keyword] {
					if !rule.Prefilter(fragment) {
						rulesToCheck[ruleID] = struct{}{}
					}
				}
			}
			putLowerBuf(lowerBufPtr)
			// Always include rules that have no keywords.
			for _, ruleID := range d.Config.NoKeywordRules {
				if !rule.Prefilter(fragment) {
					rulesToCheck[ruleID] = struct{}{}
				}
			}

			for ruleID := range rulesToCheck {
				select {
				case <-ctx.Done():
					break ScanLoop
				default:
					rule := d.Config.Rules[ruleID]
					findings = append(findings, d.detectFragmentWithRule(fragment, currentRaw, rule, encodedSegments)...)
				}
			}
```

Thoughts?

detect/detect.go
```diff
@@ -684,38 +754,40 @@ func (d *Detector) detectFragmentWithRule(fragment sources.Fragment,
...
-				continue
-			}
-		}
 
-		// check if the result matches any of the global allowlists.
-		// TODO move this to CEL-filter
-		if isAllowed, event := checkFindingAllowed(logger, finding, fragment, currentLine, d.Config.Allowlists); isAllowed {
-			event.Msg("skipping finding: global allowlist")
-			continue
+		finding.SetFingerprint()
```
bplaxco, line 779: What are your thoughts on this being an internal function on the detector?

It always feels off when there's a function that has to be called on a struct/object just to normalize its state after setting a few fields directly. 

Seems like you'd want to have something like:

```go
finding.Fingerprint = d.createFindingFingerprint(...fields...)
```

detect/detect.go
```diff
@@ -684,38 +754,40 @@ func (d *Detector) detectFragmentWithRule(fragment sources.Fragment,
...
+			}
 		}
 
-		// check if the result matches any of the rule allowlists.
-		// TODO move this to CEL-filter
-		if isAllowed, event := checkFindingAllowed(logger, finding, fragment, currentLine, r.Allowlists); isAllowed {
-			event.Msg("skipping finding: rule allowlist")
-			continue
+		// Global filter: CEL path (attributes + finding).
+		if d.Config.FilterProgram() != nil {
```
bplaxco, line 794: Not saying we should do this here or now, but I think in the long run it might make sense to have all the filter logic in it's own module/struct so it's not spread across the config and the detector 👌 

The config could potentially just hold the config and no real logic for how to deal with the findings and the detector could use it like:

```go
// Created inside the detector for backwards compatibility
NewFilter(*config.Config) (*filter.Filter, error)

// Pre regex scan filters
Filter.ShouldSkip(*config.Fragment) (bool, error)
Filter.ShouldSkipRule(*config.Rule, *sources.Fragment) (bool, error)

// Post regex scan filters
Filter.ShouldDiscard(*sources.Fragment, *detect.Finding) (bool, error)
Filter.ShouldDiscardRule(*config.Rule, *sources.Fragment, *report.Finding) (bool, error)
```

And the filter could handle the ahocorsic logic, caching etc, all in a nice optimized way.

Thoughts?

bplaxco: It might also make sense to even do the config translation logic etc, from inside the filter too. Like just make the config be a dumb data object that doesn't handle any kind of logic.

That also opens you up in the future to do seemless tweaks like having state for not rescanning files, re-reporting previous secrets, update the ignore format, etc without needing to edit the detector logic each time.

bplaxco (commented):
I wanted to stop there to get your thoughts on what's been said so far before digging too much deeper. Especially the comment about the single Filter struct to centralize the filtering logic.

cmd/git.go
zricethezav: I like it

detect/detect.go
zricethezav: > I think we should add Prefilter at the rule level too. I have rules where I skip whole directories because they're just not high enough fidelity to run on the files there.

@bplaxco I don't follow. Anything at the rule level assumes we have gotten past the ahocorasick keyword filter and regex match on some rule. `Filter` has access to all the source-level attributes so you can filter on rules.

> The rules already have a few prefilters (e.g. ahocorasick, path, etc). What are your thoughts on taking everything that isn't the ahocorasick check and regex on a rule and have translate turn that into either Prefilter or Filter?

That is already done. Ex (this could also be a `filter`):
```toml
prefilter = '''
matchesAny(attributes[?"path"].orValue(""), [
  r"""(?i)\.(?:bmp|gif|jpe?g|png|svg)$""",
  r"""(?:^|/)node_modules(?:/.*)?$""",
  r"""(?:^|/)vendor/(?:bundle|ruby)(?:/.*?)?$"""
])
||
(
  attributes[?"git.author_name"].orValue("").endsWith("[bot]") &&
  matchesAny(attributes[?"path"].orValue(""), [r"""(?:^|/)fixtures/"""])
)
'''
```

detect/detect.go
zricethezav: this version of the fingerprint will be deprecated soon. We can handle in another PR

detect/detect.go
zricethezav: I like the idea

README.md
zricethezav: yea but for the README example I dont think we should include it. Keep it simple

detect/detect.go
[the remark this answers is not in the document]
zricethezav: Changed this back to the original behavior.

README.md
```diff
@@ -66,504 +44,93 @@ cd betterleaks
...
+  } : r.status in [401, 403] ? {
+    "result": "invalid",
+    "reason": "Unauthorized"
+  } : unknown(r)
+)
+'''
+# We can override the rule's filter to add our own highly specific business logic.
+# Here we keep the baseline entropy check, but also add a contextual scenario:
+# "Ignore this GitHub PAT if it's in a mock file authored by our CI user, and contains 'TESTING'."
+filter = '''
```
rgmz: I'd put `filter` before validate so the sequence is more obvious.

scripts/release_gen.html
rgmz, line 1: What does this baby do?

docs/AGENTS.md
```diff
@@ -0,0 +1,249 @@
...
+
+---
+
+## Safe attribute access
+
+If a key is absent, a direct map access like `attributes["git.sha"]` will cause a CEL runtime error. Always use the optional accessor:
+
+```cel
+# Safe — returns "" if the key is absent
+attributes[?"git.sha"].orValue("")
```
rgmz, line 216: Random thought: it'd be nice if there was a way to return errors for attributes that _are never valid_, versus things that are known/valid but may not be present.

rgmz (commented):
I only got to 37/48 during my lunch break.

README.md
rgmz: I have no idea what happened here. It's meant to be one specific line.

scripts/release_gen.html
zricethezav: helps generate nice graphics like https://www.linkedin.com/feed/update/urn:li:activity:7448033832680914944/

docs/AGENTS.md
zricethezav: Could be something for a future `config-doctor` command

detect/detect.go
bplaxco: Ack ack.

zricethezav:
merging oh god

PR #104, opened 2026-04-23
fix(cmd): reject unknown experiments

### LINKED ISSUE #54: CLI should fail when `--experiments` contains invalid value
Things should fail, or log a warning, if an unrecognized value is specified.
```sh
$ ./betterleaks dir /tmp/foo --experiments=fake
  ○
  ○
  ●
  ○  betterleaks dev

7:21PM INF scanned ~0 bytes (0) in 50.3ms
7:21PM INF no leaks found
```

officialasishkumar:
Closes #54.

## Summary

- validate the comma-separated `--experiments` flag during command pre-run
- reject unknown experiment names instead of silently ignoring them
- keep empty experiment values accepted since no experiments are currently supported after validation was promoted

## Testing

- `go test ./cmd`
- `go test ./...`
- `go test --race ./...`
- `go run . version --experiments=fake --no-banner`
- `go generate ./...`
- `go build ./...`

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -60,6 +60,7 @@ var (
 			}
 
 			if err := validateExperiments(mustGetStringFlag(cmd, "experiments")); err != nil {
+				cmd.SilenceUsage = true
 				return err
 			}
 			return nil
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -58,6 +58,10 @@ var (
 				cmd.SetContext(ctx)
 				cobra.OnFinalize(cancel)
 			}
+
+			if err := validateExperiments(mustGetStringFlag(cmd, "experiments")); err != nil {
+				return err
+			}
 			return nil
 		},
 	}
@@ -66,6 +70,8 @@ var (
 	// and stopped after a scan completes
 	diagnosticsManager *DiagnosticsManager
 )
+
+var supportedExperiments = map[string]struct{}{}
 
 const (
 	BYTE     = 1.0
@@ -545,6 +551,23 @@ func setupValidation(cmd *cobra.Command, cfg config.Config, detector *detect.Det
 			}
 		}
 	}
+}
+
+func validateExperiments(experiments string) error {
+	for experiment := range strings.SplitSeq(experiments, ",") {
+		experiment = strings.ToLower(strings.TrimSpace(experiment))
+		if experiment == "" {
+			continue
+		}
+		if _, ok := supportedExperiments[experiment]; ok {
+			continue
+		}
+		if len(supportedExperiments) == 0 {
+			return fmt.Errorf("unknown experiment %q (no experiments are currently supported)", experiment)
+		}
+		return fmt.Errorf("unknown experiment %q", experiment)
+	}
+	return nil
 }
 
 func bytesConvert(bytes uint64) string {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -71,6 +71,7 @@ var (
 	diagnosticsManager *DiagnosticsManager
 )
 
+// Add experimental feature flag names here as they are introduced.
 var supportedExperiments = map[string]struct{}{}
 
 const (
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -1,6 +1,8 @@
 package cmd
 
 import (
+	"bytes"
+	"os"
 	"strings"
 	"testing"
 )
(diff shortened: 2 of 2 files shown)


cmd/root.go
```diff
@@ -67,6 +71,8 @@ var (
 	diagnosticsManager *DiagnosticsManager
 )
 
+var supportedExperiments = map[string]struct{}{}
```
zricethezav, line 76: Needs a comment signaling to the developer this is where experimental feature flags are added

zricethezav (changes requested):

cmd/root.go
officialasishkumar: Added a short comment above `supportedExperiments` to mark it as the place for future experimental flags. Commit: e51f761

cmd/root.go
```diff
@@ -547,6 +554,23 @@ func setupValidation(cmd *cobra.Command, cfg config.Config, detector *detect.Det
...
+	for experiment := range strings.SplitSeq(experiments, ",") {
+		experiment = strings.ToLower(strings.TrimSpace(experiment))
+		if experiment == "" {
+			continue
+		}
+		if _, ok := supportedExperiments[experiment]; ok {
+			continue
+		}
+		if len(supportedExperiments) == 0 {
+			return fmt.Errorf("unknown experiment %q (no experiments are currently supported)", experiment)
```
zricethezav, line 568: pulled this branch down, ran it with an invalid experiment:
```
/code/betterleaks/betterleaks (fix/invalid-experiments) ./betterleaks dir . --experiments=fewaf .
Error: unknown experiment "fewaf" (no experiments are currently supported)
Usage:
  betterleaks dir [flags] [path...]

Aliases:
  dir, file, directory

Flags:
      --follow-symlinks   scan files that are symlinks to other files
  -h, --help              help for dir

Global Flags:
  -b, --baseline-path string          path to baseline with issues that can be ignored
  -c, --config string                 config file path
                                      order of precedence:
                                      1. --config/-c
                                      2. env var BETTERLEAKS_CONFIG or GITLEAKS_CONFIG
                                      3. env var BETTERLEAKS_CONFIG_TOML or GITLEAKS_CONFIG_TOML with the file content
                                      4. (target path)/.betterleaks.toml or .gitleaks.toml
                                      If none of the four options are used, then the default config will be used
      --diagnostics string            enable diagnostics (http OR comma-separated list: cpu,mem,trace). cpu=CPU prof, mem=memory prof, trace=exec tracing, http=serve via net/http/pprof
      --diagnostics-dir string        directory to store diagnostics output files when not using http mode (defaults to current directory)
      --enable-rule strings           only enable specific rules by id
      --exit-code int                 exit code when leaks have been encountered (default 1)
      --experiments string            comma-separated list of experimental features to enable
  -i, --gitleaks-ignore-path string   path to .betterleaksignore or .gitleaksignore file or folder containing one (default ".")
      --ignore-gitleaks-allow         ignore gitleaks:allow and betterleaks:allow comments
  -l, --log-level string              log level (trace, debug, info, warn, error, fatal) (default "info")
      --match-context string          context around match: L (lines), C (columns/characters). e.g. 10L, 100C, -2C,+4C
      --max-archive-depth int         allow scanning into nested archives up to this depth (default "0", no archive traversal is done)
      --max-decode-depth int          allow recursive decoding up to this depth (default 5)
      --max-target-megabytes int      files larger than this will be skipped
      --no-banner                     suppress banner
      --no-color                      turn off color for verbose output
      --redact uint[=100]             redact secrets from logs and stdout. To redact only parts of the secret just apply a percent value from 0..100. For example --redact=20 (default 100%)
      --regex-engine string           regex engine (stdlib, re2) (default "re2")
  -f, --report-format string          output format (json, csv, junit, sarif, template)
  -r, --report-path string            report file (use "-" for stdout)
      --report-template string        template file used to generate the report (implies --report-format=template)
      --timeout int                   set a timeout for gitleaks commands in seconds (default "0", no timeout is set)
      --validation                    enable validation of findings against live APIs
      --validation-debug              include raw HTTP response in validation output
      --validation-extract-empty      include empty values from extractors in output
      --validation-status string      comma-separated list of validation statuses to include: valid, invalid, revoked, error, unknown, none (none = rules without validation)
      --validation-timeout duration   per-request timeout for validation (default 10s)
      --validation-workers int        number of concurrent validation workers (default 10)
  -v, --verbose                       show verbose output from scan

1:45PM FTL unknown experiment "fewaf" (no experiments are currently supported)
```

Not the prettiest. We should silence the help message here.

officialasishkumar: Addressed in 7aaa9172d5df47c6e60d34ad6e449a95c9d13fff. Invalid experiment values now suppress the Cobra usage block.

PR #110, opened 2026-04-30
feat: add rule for Selectel static API tokens

PachkaKofe04:
Added a new rule for Selectel static tokens. Selectel is a major cloud provider in the CIS region, and these tokens are frequently found in infrastructure configurations.


PachkaKofe04:
Updated the PR following the bot's feedback:
1) Moved rules from ".betterleaks.toml" to the proper code-generation pipeline
2) Created "cmd/generate/config/rules/cis.go" with 3 rules: Selectel static token, VK service token, Sber GigaChat key
3) Registered all rules in "cmd/generate/config/main.go"
4) Added entropy filters (3.5) and true/false positive test cases for each rule

PR #111, opened 2026-04-30
fix: log JSON parse error in HTTP response handler

PachkaKofe04:
When an HTTP validation response cannot be parsed as JSON, the error
was silently discarded, making it impossible to diagnose why `r.json`
returns an empty object in CEL validation rules.

Adds a debug-level log with the error and HTTP status code so rule
authors can understand what went wrong during validation.

### DIFF
--- a/celenv/bindings_http.go
+++ b/celenv/bindings_http.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


rgmz (approved):
Makes sense to me! Thoughts, @zricethezav ?

zricethezav:
thanks for the PR @PachkaKofe04

PR #119, opened 2026-05-11
Fix archive misidentification issue

bplaxco:
mholt/archives v0.1.5 was picking up .target files as .tar files. The latest version doesn't have this issue.

Also:

- Fix log message wording
- Run `make format`

```sh
# same file with different names
$ sha256sum files/*
16bd7b4b638e03010a78787aa2d5c60050021cc0fc7ebf63c8d6e83791420782  files/basic.target
16bd7b4b638e03010a78787aa2d5c60050021cc0fc7ebf63c8d6e83791420782  files/foo

# Do a scan
$ ./betterleaks --max-archive-depth=8 dir files
2:41PM INF scanned ~1704 bytes (1.70 KB) in 201ms
2:41PM WRN leaks found: 1

# Update dep
$ go get -u github.com/mholt/archives@ef71b7a32faec3ace12ec17e06eafe059ee6083e &> /dev/null
$ make build

# Try again
$ ./betterleaks --max-archive-depth=8 dir files
2:47PM INF scanned ~3408 bytes (3.41 KB) in 192ms
2:47PM WRN leaks found: 2
```

PR #121, opened 2026-05-11
Archive continue on err

bplaxco:
New behavior && should be reviewed after https://github.com/betterleaks/betterleaks/pull/119

Don't return fragments errors on archive errors
    
When Fragments() returns an error thats meant to be like
WalkDir where it signals the caller to stop traversing the
source. When an archive encounters an error it should just
log it and keep on trying things.

### DIFF
--- a/sources/file.go
+++ b/sources/file.go
@@ -125,10 +125,12 @@ func (s *File) Fragments(ctx context.Context, yield FragmentsFunc) error {
 			return nil
 		}
 		if extractor, ok := format.(archives.Extractor); ok {
-			return s.extractorFragments(ctx, extractor, stream, yield)
+			s.extractorFragments(ctx, extractor, stream, yield)
+			return nil
 		}
 		if decompressor, ok := format.(archives.Decompressor); ok {
-			return s.decompressorFragments(ctx, decompressor, stream, yield)
+			s.decompressorFragments(ctx, decompressor, stream, yield)
+			return nil
 		}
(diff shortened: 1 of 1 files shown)


sources/file.go
```diff
@@ -192,30 +195,33 @@ func (s *File) extractorFragments(ctx context.Context, extractor archives.Extrac
 		}
 
 		if err := file.Fragments(ctx, yield); err != nil {
-			return err
+			logging.Error().Err(err).Str("path", s.FullPath()).Msg("error generating file fragments")
 		}
 
 		return nil
```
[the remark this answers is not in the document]
bplaxco: Fair point. Tweaking.

sources/file.go
```diff
@@ -191,31 +194,30 @@ func (s *File) extractorFragments(ctx context.Context, extractor archives.Extrac
...
 
-		if err := file.Fragments(ctx, yield); err != nil {
-			return err
-		}
-
-		return nil
+		return file.Fragments(ctx, yield)
 	})
+
+	if err != nil {
```
bplaxco, line 200: It'd probably be ideal to add a check here to see if the error came from `file.Fragments(ctx, yield)` above or if it came from extractor.Extract.

That way you still end up with giving file.Fragments the ability to abort the traversal if it needs to.

Could do that by wrapping the error there with a special error type and then checking that at the end.

bplaxco: If we want to do that, just let me know.

zricethezav: logging the error is good for now

PR #122, opened 2026-05-12
Feat/s3 source

zricethezav:
This pull request adds support for scanning S3 buckets (and S3-compatible object stores like Cloudflare R2, MinIO, etc.) for secrets. It introduces a new `s3` command, updates documentation to describe its usage and options, and adds new resource attributes for S3 objects.

Ex:
```
./betterleaks s3 https://commoncrawl.s3.amazonaws.com/crawl-data/CC-MAIN-2017-04/warc.paths.gz --max-archive-depth=5
```

### DIFF
--- a/cmd/s3.go
+++ b/cmd/s3.go
@@ -1,0 +1,118 @@
+package cmd
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/betterleaks/betterleaks/logging"
+	"github.com/betterleaks/betterleaks/report"
+	"github.com/betterleaks/betterleaks/sources"
+)
+
+func init() {
+	rootCmd.AddCommand(s3Cmd)
+	s3Cmd.Flags().String("region", "", "AWS region (required for some non-AWS endpoints; auto-probed for AWS)")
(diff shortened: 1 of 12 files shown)


cmd/s3.go
```diff
@@ -0,0 +1,121 @@
...
+	findingSummaryAndExit(detector, findings, exitCode, start, scanErr)
+}
+
+func mustGetInt64Flag(cmd *cobra.Command, name string) int64 {
+	value, err := cmd.Flags().GetInt64(name)
+	if err != nil {
+		logging.Fatal().Err(err).Msgf("could not get flag: %s", name)
+	}
+	return value
+}
```
bplaxco: Would it make sense to put all the `mustGet*`'s in a single file?

bplaxco (commented):
Haven't dug through all the code, but I like the general  interface for the command :ok_hand:

cmd/s3.go
zricethezav: Probably!

zricethezav:
@bplaxco @rgmz should we support bucket enumeration like this PR proposes?

zricethezav:
merging

PR #123, opened 2026-05-12
more generic agents md

### LINKED ISSUE #117: Write a more generic agents.md
with less markdown headers :barf:

zricethezav:
@rgmz @bplaxco feel free to add to this. Should be lightweight


rgmz:
What's the motivation behind 'generic' here? While the current AGENTS.md could be trimmed, it'd be a mistake to strip out all the useful information. Most models prefer specificity; alternatively, we could use skills or link to another file in docs/ or CONTRIBUTING.md.

The current file is *too* generic to be useful. At a minim, we should have basic commands and lifecycle for development.

zricethezav:
removing agents.md

PR #125, opened 2026-05-14
Use strconv.Quote for quoting CEL strings

bplaxco:
# Error

Some regex patterns weren't being escaped properly. 

I believe this is an example of narrowing it down:

```toml
[[rules]]
id = 'someid'
description = 'Example'
regex = '''example'''
[rules.allowlist]
regexes = [
# This line freaks out the CEL compiler
'''(?i)<\w+\s[^>]{0,128}token\w*\s*=\s*\"@?[a-z.\/\-]+\"''',
]
```

# The Fix

> ⚠️ **WARNING: I've only done light testing with this so far.**
>
> I wanted to go ahead and submit this PR so folks could take a look at it and test it more thoroughly 
> (esp those who are more familiar with CEL string requirements).

I swapped out the custom quoting function with `strconv.Quote`. It seems to handle the edge cases pretty well.

# Reproducing the error

```sh
#!/bin/sh
set -x

workdir="$(mktemp -d)"
trap 'rm -rf "${workdir}"' exit

# Setup betterleaks
git clone --revision 8552eb997621298f5e5fb2b73e3bf64b872dd9ea --depth 1 https://github.com/betterleaks/betterleaks.git "${workdir}/betterleaks"
(cd "${workdir}/betterleaks" && make build)

# Setup patterns
curl -Lo "${workdir}/betterleaks.toml" https://raw.githubusercontent.com/leaktk/patterns/e2a5cb903115039cce42de3be1947402f10207c6/target/patterns/gitleaks/8.27.0

# Setup test-repo
git clone --revision f15521a2b100276d598410a9e1e46a11a397ad7e --depth 1 https://github.com/leaktk/fake-leaks.git  "${workdir}/fake-leaks"

# Run scan
"${workdir}/betterleaks/betterleaks" \
    --config "${workdir}/betterleaks.toml" \
    git "${workdir}/fake-leaks"
```


bplaxco:
Here's a condensed version of the original problem:


```go
package main

import (
	"fmt"
	"strings"
)

func celStringLit(s string) string {
	// If it contains a backslash AND doesn't contain a triple-quote, raw strings are safe.
	if strings.ContainsRune(s, '\\') && !strings.Contains(s, `"""`) {
		return `r"""` + s + `"""`
	}

	var b strings.Builder
	b.WriteByte('"')
	for i := 0; i < len(s); i++ {
		c := s[i]
		switch c {
		case '"':
			b.WriteString(`\"`)
		case '\\':
			b.WriteString(`\\`)
		case '\n':
			b.WriteString(`\n`)
		case '\r':
			b.WriteString(`\r`)
		case '\t':
			b.WriteString(`\t`)
		default:
			b.WriteByte(c)
		}
	}
	b.WriteByte('"')
	return b.String()
}


func main() {
	str := `(?i)<\w+\s[^>]{0,128}token\w*\s*=\s*\"@?[a-z.\/\-]+\"`
	fmt.Println(celStringLit(str))
}
```

It outputs:

```CEL
r"""(?i)<\w+\s[^>]{0,128}token\w*\s*=\s*\"@?[a-z.\/\-]+\""""
```

If you run that in a CEL playground like this one: https://playcel.undistro.io/

You get an error like:

```
failed to compile the CEL expression: ERROR: <input>:8:60: Syntax error: token recognition error at: '"'
 | r"""(?i)<\w+\s[^>]{0,128}token\w*\s*=\s*\"@?[a-z.\/\-]+\""""
 | ...........................................................^
 ```
 
 But with `strconv.Quote` you get:
 
 ```CEL
 "(?i)<\\w+\\s[^>]{0,128}token\\w*\\s*=\\s*\\\"@?[a-z.\\/\\-]+\\\""
 ```
 
 Which is valid CEL :ok_hand:

**Comment by zricethezav** [comment]
@bplaxco in theory I approve this PR, but I'd like to migrate all the rule generation code to `filter` and `prefilter` first before merging this.

The reason being is this (after running go generate ./...):
```diff
diff --git a/config/betterleaks.toml b/config/betterleaks.toml
index a825869..b5cda3a 100644
--- a/config/betterleaks.toml
+++ b/config/betterleaks.toml
@@ -22,49 +22,49 @@ betterleaksMinVersion = "v1.2.0"

 prefilter = '''
 matchesAny(attributes[?"path"].orValue(""), [
-  r"""gitleaks\.toml""",
-  r"""(?i)\.(?:bmp|gif|jpe?g|png|svg|tiff?)$""",
-  r"""(?i)\.(?:eot|[ot]tf|woff2?)$""",
-  r"""(?i)\.(?:docx?|xlsx?|pdf|bin|socket|vsidx|v2|suo|wsuo|.dll|pdb|exe|gltf)$""",
-  r"""go\.(?:mod|sum|work(?:\.sum)?)$""",
-  r"""(?:^|/)vendor/modules\.txt$""",
-  r"""(?:^|/)vendor/(?:github\.com|golang\.org/x|google\.golang\.org|gopkg\.in|istio\.io|k8s\.io|sigs\.k8s\.io)(?:/.*)?$""",
-  r"""(?:^|/)gradlew(?:\.bat)?$""",
-  r"""(?:^|/)gradle\.lockfile$""",
-  r"""(?:^|/)mvnw(?:\.cmd)?$""",
-  r"""(?:^|/)\.mvn/wrapper/MavenWrapperDownloader\.java$""",
+  "gitleaks\\.toml",
+  "(?i)\\.(?:bmp|gif|jpe?g|png|svg|tiff?)$",
+  "(?i)\\.(?:eot|[ot]tf|woff2?)$",
+  "(?i)\\.(?:docx?|xlsx?|pdf|bin|socket|vsidx|v2|suo|wsuo|.dll|pdb|exe|gltf)$",
+  "go\\.(?:mod|sum|work(?:\\.sum)?)$",
+  "(?:^|/)vendor/modules\\.txt$",
+  "(?:^|/)vendor/(?:github\\.com|golang\\.org/x|google\\.golang\\.org|gopkg\\.in|istio\\.io|k8s\\.io|sigs\\.k8s\\.io)(?:/.*)?$",
+  "(?:^|/)gradlew(?:\\.bat)?$",
+  "(?:^|/)gradle\\.lockfile$",
+  "(?:^|/)mvnw(?:\\.cmd)?$",
+  "(?:^|/)\\.mvn/wrapper/MavenWrapperDownloader\\.java$",
```

`(?:^|/)vendor/(?:github\\.com|golang\\.org/x|google\\.golang\\.org|gopkg\\.in|istio\\.io|k8s\\.io|sigs\\.k8s\\.io)(?:/.*)?$`,

specifically `github\\.com`. Those double slashes 😬 . It get translated in CEL correctly but if you are just copying this from the config, it'd be wrong.

bplaxco:
>  It get translated in CEL correctly but if you are just copying this from the config, it'd be wrong.

@zricethezav I _think_ I follow but could you walk me through that concern a bit more?

zricethezav:
`gitleaks\\.toml` != `gitleaks\.toml`

When users open their configs they should be able to copy and interact with the pattern the way the underlying engine uses it.

strconv converts to `"gitleaks\\.toml"` which gets correctly parsed in toml -> cel as `gitleaks\.toml` but it's confusing if youre just looking at the config

bplaxco:
Ack ack. I'll see if I can take a stab at implementing it with the `r"""` syntax but still handle that trailing `"` at the end.

zricethezav:
@bplaxco actually strconv.Quote might work fine since this is for translating. We just need to convert all the allowlists in the rule generator

zricethezav:
@bplaxco good to close since we merged https://github.com/betterleaks/betterleaks/pull/145/changes#diff-65c9d3d95e7527dcef400fdfccd2ef259fbbfb620ea9e911e1e694e343f814d9R182. Reopen if I'm wrong

PR #127, opened 2026-05-14
fix: Remove stray comma from nytimes identifier

PachkaKofe04:
The 'new-york-times,' identifier in the 'GenerateSemiGenericRegex' call
had a trailing comma inside the string literal, causing the regex to
match 'new-york-times,' (with comma) instead of 'new-york-times'.
The 'Keywords' slice was correct - only the regex identifier was affected.


rgmz (approved):

rgmz:
Thank you!

PR #129, opened 2026-05-14
Add ValidateCEL for Stripe and Datadog rules

PachkaKofe04:
**Stripe** ('stripe-access-token')
- Endpoint: 'GET https://api.stripe.com/v1/account'
- Auth: 'Authorization: Bearer {token}'
- Valid when: status 200 and 'object == "account"'

**Datadog** ('datadog-access-token')
- Endpoint: 'GET https://api.datadoghq.com/api/v1/validate'
- Auth: 'DD-API-KEY: {token}' header
- Valid when: status 200 and 'valid == true'

Both expressions follow the existing 'cel.bind' pattern used by other rules (OpenAI, Anthropic, etc.) and use 'finding["secret"]' per the 'celenv/validation.go' convention.

'config/betterleaks.toml' regenerated via 'go generate ./...'.

### DIFF
--- a/cmd/generate/config/rules/datadog.go
+++ b/cmd/generate/config/rules/datadog.go
@@ -16,6 +16,17 @@ func DatadogtokenAccessToken() *config.Rule {
 		Keywords: []string{
 			"datadog",
 		},
+		ValidateCEL: `cel.bind(r,
+  http.get("https://api.datadoghq.com/api/v1/validate", {
+    "DD-API-KEY": finding["secret"]
+  }),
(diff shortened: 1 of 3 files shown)


PachkaKofe04:
Fixed the Stripe validator based on the review feedback:
Switched endpoint from /v1/account to /v1/balance (broader permission coverage for restricted keys)
Separated 401 and 403 handling: 401 → invalid (bad key), 403 → valid with "Restricted key" reason, since Stripe returns 403 when the key is valid but lacks permission for the endpoint - not when the key itself is wrong

zricethezav:
no longer using CEL

PR #144, opened 2026-05-21
readme: fix link to betterleaks default config

becojo:
It's using the `master` brain instead of `main`.


zricethezav:
Great catch!

PR #146, opened 2026-05-22
use internal for pkgs that shouldnt be exposed

zricethezav:


zricethezav:
@twpayne This wont break anything on your end right? Looking at the code it looks like youre just using `"github.com/betterleaks/betterleaks/detect"`

twpayne:
Thank you for checking! Yup, as you say, chezmoi only imports `github.com/betterleaks/betterleaks/detect` so this is fine :)

rgmz (approved):

PR #155, opened 2026-06-01
feat(dir): include/exclude globs and path patterns in ignore files

### LINKED ISSUE #150: ignore files and directories
Is there any way to ignore the whole directory or at least file? I have a directory with many test keys (about 5k) and want to ignore it. 

It seems that it can be done with config, but it is static in my case and only .betterleaksignore can be changed

shahravir:
## Summary

- Add `--include` and `--exclude` path glob flags to `betterleaks dir` (aligned with maintainer feedback on #150).
- Allow `.betterleaksignore` / `.gitleaksignore` to list **path globs** in addition to finding fingerprints, so repos can skip directories (e.g. test key fixtures) without changing a shared Docker image or CLI args.

Fixes #150

## Motivation

Issue #150 asked for a way to ignore whole directories when scanning. Maintainers suggested `--include`/`--exclude` on the directory command; reporters also need repository-local rules via `.betterleaksignore` when the launch command is fixed org-wide.

## Changes

| Area | What |
|------|------|
| `cmd/directory.go` | `--include`, `--exclude` flags |
| `sources/pathglob.go` | Glob matching (`**`, trailing `/`) |
| `sources/files.go` | Skip paths during directory walk |
| `detect/detect.go` | Parse non-fingerprint lines in ignore files as path excludes |
| `docs/scanning.md` | Usage examples |

## Test plan

- [x] `go test ./sources/... -run TestShouldScanPath`
- [x] `go test ./detect/... -run TestAddGitleaksIgnore_pathGlobs`
- [ ] Manual: `betterleaks dir . --exclude 'fixtures/**'`
- [ ] Manual: add `fixtures/test-keys/**` to `.betterleaksignore` and confirm directory is skipped

## Example

```sh
betterleaks dir . --exclude 'fixtures/test-keys/**'
```

```gitignore
# .betterleaksignore
fixtures/test-keys/**
api/api.go:aws-access-key:7
```

### DIFF
--- a/cmd/directory.go
+++ b/cmd/directory.go
@@ -16,6 +16,8 @@ import (
 func init() {
 	rootCmd.AddCommand(directoryCmd)
 	directoryCmd.Flags().Bool("follow-symlinks", false, "scan files that are symlinks to other files")
+	directoryCmd.Flags().StringSlice("include", []string{}, "only scan files matching these path globs")
+	directoryCmd.Flags().StringSlice("exclude", []string{}, "skip files and directories matching these path globs")
 }
 
 var directoryCmd = &cobra.Command{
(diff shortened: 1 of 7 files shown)


shahravir:
Thanks @greptile-apps — addressed both review points in 5739fb3:

**P1 — include filter skipped walk root:** Added `ShouldWalkDir` so directories that are *ancestors* of an include pattern (e.g. `.` with `src/**`) are still entered. `ShouldScanPath` remains for files only.

**P2 — basename fallback:** Removed the implicit `filepath.Base` retry; globs now match the full path only. Docs note using `**/*.go` instead of `*.go` for extension-wide excludes.

Added `TestShouldWalkDir` including the `.` + `src/**` case.

PR #174, opened 2026-06-08
Make ValidationStatus a typed enum

### LINKED ISSUE #55: Make ValidationStatus an enum (define once, use everywhere type of deal)
If we make `Finding.ValidationStatus` an 'enum', it can be reused here for safety.

_Originally posted by @rgmz in https://github.com/betterleaks/betterleaks/pull/53#discussion_r2956723117_

mazen-salah:
`Finding.ValidationStatus` (and `RequiredSet.ValidationStatus`) were bare strings, with the canonical values (`valid`, `invalid`, `revoked`, …) spelled out as literals across the validation, detection, and reporting code — error-prone, as noted in #53.

This introduces a string-backed `report.ValidationStatus` type with named constants and threads it through the validation pipeline: `validate.Result.Status`, `BetterStatus`, the recognised/priority tables, and the detector's `ValidationCounts`. Because the type is string-backed it serializes exactly as before, so JSON/SARIF reports are unchanged (locked down by a new test).

Closes #55


zricethezav (approved):

report/validation_status_test.go
```diff
@@ -0,0 +1,35 @@
+package report
```
zricethezav: @mazen-salah We really don't need a test for this. remove this file

mazen-salah: @zricethezav done, removed. Thanks!

PR #176, opened 2026-06-08
Fix redaction of capture groups and multi-byte values

mazen-salah:
Two redaction bugs in `report.Finding`:

- `Redact()` masked Line/Match/MatchContext/RequiredSets but never touched `CaptureGroups`, which can hold the captured value verbatim and are emitted in the JSON, JUnit, and template reports — so `--redact` still leaked it there. Capture-group values are now redacted with the same replacement.
- `MaskSecret()` measured length and sliced by byte offset, which splits a rune on multi-byte UTF-8 values (yielding invalid UTF-8) and computes the mask ratio from byte count rather than character count. It now operates on runes. ASCII behavior is unchanged.

Added tests covering both.


report/redaction_test.go
```diff
@@ -0,0 +1,51 @@
+package report
```
zricethezav: @mazen-salah delete file. Move test to finding_test.go

mazen-salah: @zricethezav done, moved them into finding_test.go and deleted the file. Sorry for the noise, thanks!

PR #177, opened 2026-06-08
dir: surface scan errors so a failed scan exits non-zero

mazen-salah:
`runDirectory` collected per-file scan errors only to log them; the `scanErr` passed to `findingSummaryAndExit` was always nil, so `dir` reported a clean run and exited 0 even when files could not be scanned.

This accumulates the errors and wraps them in `multipleErrors` (the same pattern the `git` command already uses) so a partial or failed scan is reported and exits non-zero.

### DIFF
--- a/cmd/directory.go
+++ b/cmd/directory.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"fmt"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


zricethezav (approved):
reasonable. Should probably include this for all sources

PR #178, opened 2026-06-08
detect: make config/baseline self-scan guard separator-insensitive

mazen-salah:
The file source normalizes fragment paths to forward slashes (`filepath.ToSlash`), but `Config.Path` and `baselinePath` keep the native separator (`filepath.Join`). On Windows the two never compared equal, so the config and baseline files were scanned against their own rule patterns and produced spurious findings.

This compares the paths through a separator-normalizing helper instead of raw string equality. Added tests.


zricethezav (approved):

detect/detect.go
```diff
@@ -558,10 +559,19 @@ func (d *Detector) DetectString(content string) []report.Finding {
 	})
 }
 
+// samePath reports whether two file paths refer to the same location, tolerating
+// OS separator differences. The file source normalizes fragment paths to forward
+// slashes (filepath.ToSlash), whereas config/baseline paths keep the native
+// separator, so a raw == comparison misses on Windows and the config or baseline
+// file ends up being scanned against itself.
+func samePath(a, b string) bool {
```
zricethezav: @mazen-salah move to utils.go. Delete the test file. use utils_test.go

zricethezav (changes requested):

detect/detect.go
mazen-salah: @zricethezav done, moved samePath into utils.go and the test into utils_test.go. Thanks!

PR #186, opened 2026-06-08
fix(curl-auth-user): filter nested bash parameter expansions as false positives

antoineauger:
Hello there,

Thanks a lot for providing betterleaks ❤️ 

We just noticed a small bug that affects us: the existing filter for shell variable references only handles simple patterns like `$VAR` or `${VAR}`. 

Bash parameter expansions with defaults (e.g. `${USER:-${FALLBACK}}`) contain colons and nested braces that break
the existing pattern, causing false positives for lines like:

```
  curl -u "${USER:-${DEFAULT_USER}}:${APIKEY:-${DEFAULT_APIKEY}}"
```

This PR adds a new CEL filter regex that matches when both sides of the
`user:password` separator are `${...}` expressions with up to one level
of nested braces, covering `:-`, `:+`, and `:=` expansion operators.

🛠️ with ❤️ by [Siemens](https://opensource.siemens.com/)


config/betterleaks.toml
```diff
@@ -4793,7 +4794,6 @@ entropy(finding["secret"]) <= 2.0
 id = "scaleway-secret-key"
 description = "Identified a standalone Scaleway Secret Key. This can be used to authenticate API requests."
 regex = '''(?i)[\w.-]{0,50}?(?:(?:scaleway|scw).{0,20}?(?:secret|token))(?:[ \t\w.-]{0,20})[\s'"]{0,3}(?:=|>|:{1,3}=|\|\||:|=>|\?=|,)[\x60'"\s=]{0,5}([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})(?:\\?['"\x60]|[\s;]|\\[nr]|$)'''
-entropy = 3
```
antoineauger, line 4796: Not related to the changes in this PR but this was deleted when I ran `make config/betterleaks.toml` 😁

antoineauger:
WDYT of this new addition @zricethezav? Some of our users are complaining about "curl-auth-user" false positives. 

I asked some help to Claude here (see prompt below) and then ran the tests/the generate commands.  Happy to rework that if you have some feedback.

Thanks a lot 😸 

/cc @bufferoverflow

---

Claude prompt:

> Because of the `curl-auth-user` rule, this false positive is detected: 
> `curl -u "${USER:-${DEFAULT_USER}}:${APIKEY:-${DEFAULT_APIKEY}}"`
> How can I change that?

zricethezav:
Hey! An actual human! 😅

I think this is a great change @antoineauger and I really appreciate Siemens using Betterleaks. You guys have helped surface plenty of worthwhile bugs over the years.

PR #188, opened 2026-06-08
Match --exclude-repo globs OS-independently and case-insensitively

mazen-salah:
`isExcluded` used `filepath.Match`, whose separator handling is OS-dependent, so `--exclude-repo` globs containing `/` matched differently on Windows than on Linux. GitHub matching was also case-sensitive despite GitHub names being case-insensitive.

Switch the GitHub and GitLab matchers to `path.Match` (always `/`-separated) and case-fold both pattern and name. Adds tests.


sources/exclude_repo_test.go
```diff
@@ -0,0 +1,36 @@
+package sources
```
zricethezav: @mazen-salah remove test

mazen-salah: @zricethezav done, removed 🙏

PR #190, opened 2026-06-08
Surface the real git stderr instead of a generic error

mazen-salah:
When git wrote to stderr, the scan failed with `errors.New("stderr is not empty")`, so callers and the scan summary lost the actual cause — e.g. a submodule scan reported only `stderr is not empty` instead of the real `fatal: not a git repository: ...`.

Collect the offending stderr lines and include them in the returned error; benign warnings (rename-limit, auto-pack, gc) are still ignored. Adds tests.


sources/stderr_test.go
```diff
@@ -0,0 +1,37 @@
+package sources
```
zricethezav: @mazen-salah remove test. Stop submitting PRs w/ extraneous test files

mazen-salah: @zricethezav sorry about that, removed the test. I will stop adding standalone test files going forward. Thanks!

mazen-salah:
@zricethezav just to be clear, these aren't auto-generated spam. I'm using Claude Code to help me find edge cases and real bugs in the codebase, and I read, test, and review every change myself before opening a PR. Happy to slow down or adjust the format to whatever is least noisy for you.

PR #196, opened 2026-06-09
Rules/more rules 3

zricethezav:


zricethezav:
"`Adds time.nowRFC3339; correct.` "

so authoritative lol

PR #207, opened 2026-06-16
Filter Publishable and Test Keys from Generic Results

jl-aikido:
This PR adds support for a generics filter that removes candidate secrets of known public-by-design and test keys. The most obvious example is a public Stripe key, but there are many other in this list. 

Some patterns rely on both a regex and a keyword. Even with the keyword, it's possible some of these are too broad (e.g. `[r"""^[0-9a-zA-Z_-]{20}$"""])`); however, the keyword match is based on the line, so it's quite contained. Open to feedback on the implementation and any particular "detectors".


cmd/generate/config/rules/generic_api_key_filter.go
```diff
@@ -25,7 +25,116 @@ func buildGenericAPIKeyFilter() string {
...
+// also appears nearby on the finding's line.
+type testAndPublicAPIFilter struct {
+	regex    string
+	keywords []string
+}
+
+const uuidHex = `[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}`
+
+// testAndPublicAPIFilters is a denylist of test and public API key patterns.
+var testAndPublicAPIFilters = []testAndPublicAPIFilter{
```
rgmz, line 42: Do we have any examples we could use as test cases?

config/betterleaks.toml
```diff
@@ -3856,6 +3856,73 @@ entropy(finding["secret"]) <= 3.5
...
+|| matchesAny(finding["secret"], [r"""^public-token-live-[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$"""])
+|| matchesAny(finding["secret"], [r"""^pdl_sdbx_apikey_[a-z\d]{26}_[a-zA-Z\d]{22}_[a-zA-Z\d]{3}$"""])
+|| (matchesAny(finding["secret"], [r"""^pk_test_[A-Za-z0-9]{24}(?:[A-Za-z0-9]{10})?(?:[A-Za-z0-9]{65})?$"""]) && containsAny(finding["line"], ["stripe", "woo", "wcm"]))
+|| (matchesAny(finding["secret"], [r"""^rk_test_[A-Za-z0-9]{24}(?:[A-Za-z0-9]{10})?(?:[A-Za-z0-9]{65})?$"""]) && containsAny(finding["line"], ["stripe"]))
+|| (matchesAny(finding["secret"], [r"""^sk_test_[A-Za-z0-9]{24}(?:[A-Za-z0-9]{10})?(?:[A-Za-z0-9]{65})?$"""]) && containsAny(finding["line"], ["stripe", "woo", "wcm"]))
+|| (matchesAny(finding["secret"], [r"""^pk_live_[A-Za-z0-9]{24}(?:[A-Za-z0-9]{10})?(?:[A-Za-z0-9]{65})?$"""]) && containsAny(finding["line"], ["stripe", "woo", "wcm"]))
+|| (matchesAny(finding["secret"], [r"""^pk_test_[A-Za-z0-9_-]{40}$"""]) && containsAny(finding["line"], ["paystack"]))
+|| (matchesAny(finding["secret"], [r"""^sk_test_[A-Za-z0-9_-]{40}$"""]) && containsAny(finding["line"], ["paystack"]))
```
rgmz, line 3736: Is `containsAny` cheaper than `matchesAny`? If so, it could be worth swapping the order.

rgmz (approved):
These _should_ be good, I don't see any obvious issues; if we can include either documentation or test cases that would be great (for posterity).

config/betterleaks.toml
zricethezav: `containsAny` should be cheaper since it's ahocorasick string cmps vs a big ol pattern match

zricethezav:
@jl-aikido changes look great! The only thing that would be helpful would be for some comments in the helper describing what some of these regexes are for

PR #228, opened 2026-07-07
cuts down scan time a decent amount

zricethezav:


docs/config.md
```diff
@@ -56,6 +56,18 @@ r.json?.login ?? ""
 The full attributes source is maintained in
 [`sources/attribute.go`](https://github.com/betterleaks/betterleaks/blob/main/sources/attribute.go).
 
+Common `finding` keys:
+
+| Key | Description |
+| :--- | :--- |
+| `secret` | Captured secret value after applying capture groups or `secretGroup`. |
+| `match` | Exact regex match text. This does not include synthetic pre-match context. |
+| `match_extended` | `match` plus up to 50 bytes immediately before the match. Useful for filters that need nearby identifier/context text without making the regex more expensive. |
```
rgmz, line 65: Why only _before_ the match? What if we need/want to add something _after_ the match later? I don't think `match_extended` is a great name for this.

zricethezav: Mainly to preserve existing functionality and emulate the pattern before since we have rules that depend on content before the identifier. I even considered ‘match_50c’ if you really wanna talk about bad names

zricethezav: better idea:
```
filter.matchesAnyNearMatch(finding, [`...`], 50, 0)
filter.containsAnyNearMatch(finding, ["stripe", "woo"], 50, 0)
```

zricethezav:
closing, using a different branch

PR #230, opened 2026-07-08
validation check for existing gcp-api-key rule, create new gemini rules

zricethezav:
This pull request adds support for detecting and validating Google Gemini API keys and validating existing GCP API keys (originally designed to be public) against the gemini api

relevant https://trufflesecurity.com/blog/google-api-keys-werent-secrets-but-then-gemini-changed-the-rules

---
Conditions:

- **identitytoolkit invalid** -> `invalid`; key is not a valid Google API key.
- **Gemini 200 with models** -> `valid`; active Google key with Gemini access.
- **Gemini API_KEY_SERVICE_BLOCKED** -> `needs_validation`; active Google key, but Gemini is
  blocked.
- **Gemini referrer/IP/iOS/Android restriction** -> `needs_validation`; Gemini access unknown
  because another caller context may work.
- **identitytoolkit active/blocked + Gemini 400/API_KEY_INVALID** -> `needs_validation`;
  active Google key, but Gemini does not accept it and other Google APIs may still work.
- **Other Gemini 403** -> `needs_validation`; active Google key with unknown Gemini access.

### DIFF
--- a/cmd/generate/config/rules/gcp.go
+++ b/cmd/generate/config/rules/gcp.go
@@ -11,7 +11,9 @@ const gcpAPIKeyValidationExpr = `let k = http.get("https://www.googleapis.com/id
     "result": "invalid",
     "reason": "Unauthorized"
   } : (let r = http.get("https://generativelanguage.googleapis.com/v1beta/models?key=" + finding["secret"], {}); r.status == 200 && (r.body contains '"models"') ? {
-    "result": "valid"
+    "result": "valid",
+    "active_google_key": true,
+    "gemini_access": true
   } : r.status == 403 && (r.body contains "API_KEY_HTTP_REFERRER_BLOCKED") ? {
     "result": "needs_validation",
     "reason": "Active Google API key blocked by HTTP referrer restriction",
@@ -38,7 +40,7 @@ const gcpAPIKeyValidationExpr = `let k = http.get("https://www.googleapis.com/id
     "restriction": "android_app"
   } : r.status == 403 && (r.body contains "API_KEY_SERVICE_BLOCKED") ? {
     "result": "needs_validation",
-    "reason": "Active Google API key without Gemini access",
+    "reason": "Active Google API key blocked from Gemini API",
     "active_google_key": true,
     "gemini_access": false
   } : r.status == 403 ? {
@@ -48,9 +50,12 @@ const gcpAPIKeyValidationExpr = `let k = http.get("https://www.googleapis.com/id
     "gemini_access": "unknown"
   } : r.status == 400 && k.status in [200, 403] ? {
     "result": "needs_validation",
-    "reason": "Active Google API key without Gemini access",
+    "reason": "Active Google API key not accepted by Gemini API",
     "active_google_key": true,
     "gemini_access": false
+  } : r.status == 400 && (r.body contains "API_KEY_INVALID") ? {
+    "result": "invalid",
+    "reason": "Unauthorized"
   } : r.status == 400 ? {
     "result": "invalid",
     "reason": "Unauthorized"
--- a/cmd/generate/config/rules/gcp.go
+++ b/cmd/generate/config/rules/gcp.go
@@ -68,7 +68,18 @@ func GCPAPIKey() *config.Rule {
 		Description: "Uncovered a GCP API key, which could lead to unauthorized access to Google Cloud services and data breaches.",
 		Regex:       utils.GenerateUniqueTokenRegex(`AIza[\w-]{35}`, false),
 		Keywords:    []string{"AIza"},
-		Filter:      "entropy(finding[\"secret\"]) <= 4.0\n|| matchesAny(finding[\"secret\"], [\n  `AIzaSyabcdefghijklmnopqrstuvwxyz1234567`,\n  `AIzaSyAnLA7NfeLquW1tJFpx_eQCxoX-oo6YyIs`,\n  `AIzaSyCkEhVjf3pduRDt6d1yKOMitrUEke8agEM`,\n  `AIzaSyDMAScliyLx7F0NPDEJi1QmyCgHIAODrlU`,\n  `AIzaSyD3asb-2pEZVqMkmL6M9N6nHZRR_znhrh0`,\n  `AIzayDNSXIbFmlXbIE6mCzDLQAqITYefhixbX4A`,\n  `AIzaSyAdOS2zB6NCsk1pCdZ4-P6GBdi_UUPwX7c`,\n  `AIzaSyASWm6HmTMdYWpgMnjRBjxcQ9CKctWmLd4`,\n  `AIzaSyANUvH9H9BsUccjsu2pCmEkOPjjaXeDQgY`,\n  `AIzaSyA5_iVawFQ8ABuTZNUdcwERLJv_a_p4wtM`,\n  `AIzaSyA4UrcGxgwQFTfaI3no3t7Lt1sjmdnP5sQ`,\n  `AIzaSyDSb51JiIcB6OJpwwMicseKRhhrOq1cS7g`,\n  `AIzaSyBF2RrAIm4a0mO64EShQfqfd2AFnzAvvuU`,\n  `AIzaSyBcE-OOIbhjyR83gm4r2MFCu4MJmprNXsw`,\n  `AIzaSyB8qGxt4ec15vitgn44duC5ucxaOi4FmqE`,\n  `AIzaSyA8vmApnrHNFE0bApF4hoZ11srVL_n0nvY`\n])",
+		ValidateExpr: `let r = http.get("https://generativelanguage.googleapis.com/v1beta/models?key=" + finding["secret"], {}); r.status == 200 && (r.body contains '"models"') ? {
+    "result": "valid"
+  } : r.status == 403 ? {
+    "result": "needs_validation",
+    "reason": "Active Google API key without Gemini access",
+    "active_google_key": true,
+    "gemini_access": false
+  } : r.status == 400 ? {
+    "result": "invalid",
+    "reason": "Unauthorized"
+  } : validate.unknown(r)`,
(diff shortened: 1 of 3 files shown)


zricethezav:
@jl-aikido look good?

cmd/generate/config/rules/gcp.go
```diff
@@ -68,7 +68,18 @@ func GCPAPIKey() *config.Rule {
...
 		Regex:       utils.GenerateUniqueTokenRegex(`AIza[\w-]{35}`, false),
 		Keywords:    []string{"AIza"},
-		Filter:      "entropy(finding[\"secret\"]) <= 4.0\n|| matchesAny(finding[\"secret\"], [\n  `AIzaSyabcdefghijklmnopqrstuvwxyz1234567`,\n  `AIzaSyAnLA7NfeLquW1tJFpx_eQCxoX-oo6YyIs`,\n  `AIzaSyCkEhVjf3pduRDt6d1yKOMitrUEke8agEM`,\n  `AIzaSyDMAScliyLx7F0NPDEJi1QmyCgHIAODrlU`,\n  `AIzaSyD3asb-2pEZVqMkmL6M9N6nHZRR_znhrh0`,\n  `AIzayDNSXIbFmlXbIE6mCzDLQAqITYefhixbX4A`,\n  `AIzaSyAdOS2zB6NCsk1pCdZ4-P6GBdi_UUPwX7c`,\n  `AIzaSyASWm6HmTMdYWpgMnjRBjxcQ9CKctWmLd4`,\n  `AIzaSyANUvH9H9BsUccjsu2pCmEkOPjjaXeDQgY`,\n  `AIzaSyA5_iVawFQ8ABuTZNUdcwERLJv_a_p4wtM`,\n  `AIzaSyA4UrcGxgwQFTfaI3no3t7Lt1sjmdnP5sQ`,\n  `AIzaSyDSb51JiIcB6OJpwwMicseKRhhrOq1cS7g`,\n  `AIzaSyBF2RrAIm4a0mO64EShQfqfd2AFnzAvvuU`,\n  `AIzaSyBcE-OOIbhjyR83gm4r2MFCu4MJmprNXsw`,\n  `AIzaSyB8qGxt4ec15vitgn44duC5ucxaOi4FmqE`,\n  `AIzaSyA8vmApnrHNFE0bApF4hoZ11srVL_n0nvY`\n])",
+		ValidateExpr: `let r = http.get("https://generativelanguage.googleapis.com/v1beta/models?key=" + finding["secret"], {}); r.status == 200 && (r.body contains '"models"') ? {
+    "result": "valid"
+  } : r.status == 403 ? {
```
jl-aikido: If you hit a 403, it will mean one of two things related to Gemini: (a) it does not have access to Gemini, but instead can access some other GCP APIs, (b) it might be able to access Gemini, but the key is restricted and it requires bypassing those restrictions to figure that out. We can further parse the response to get at those nuances if we want; otherwise, we can just keep it as is?

zricethezav: > it might be able to access Gemini, but the key is restricted and it requires bypassing those restrictions to figure that out. We can further parse the response to get at those nuances if we want

I wasnt aware of that, we should definitely further parse the response

cmd/generate/config/rules/gcp.go
```diff
@@ -68,7 +68,18 @@ func GCPAPIKey() *config.Rule {
...
 		Regex:       utils.GenerateUniqueTokenRegex(`AIza[\w-]{35}`, false),
 		Keywords:    []string{"AIza"},
-		Filter:      "entropy(finding[\"secret\"]) <= 4.0\n|| matchesAny(finding[\"secret\"], [\n  `AIzaSyabcdefghijklmnopqrstuvwxyz1234567`,\n  `AIzaSyAnLA7NfeLquW1tJFpx_eQCxoX-oo6YyIs`,\n  `AIzaSyCkEhVjf3pduRDt6d1yKOMitrUEke8agEM`,\n  `AIzaSyDMAScliyLx7F0NPDEJi1QmyCgHIAODrlU`,\n  `AIzaSyD3asb-2pEZVqMkmL6M9N6nHZRR_znhrh0`,\n  `AIzayDNSXIbFmlXbIE6mCzDLQAqITYefhixbX4A`,\n  `AIzaSyAdOS2zB6NCsk1pCdZ4-P6GBdi_UUPwX7c`,\n  `AIzaSyASWm6HmTMdYWpgMnjRBjxcQ9CKctWmLd4`,\n  `AIzaSyANUvH9H9BsUccjsu2pCmEkOPjjaXeDQgY`,\n  `AIzaSyA5_iVawFQ8ABuTZNUdcwERLJv_a_p4wtM`,\n  `AIzaSyA4UrcGxgwQFTfaI3no3t7Lt1sjmdnP5sQ`,\n  `AIzaSyDSb51JiIcB6OJpwwMicseKRhhrOq1cS7g`,\n  `AIzaSyBF2RrAIm4a0mO64EShQfqfd2AFnzAvvuU`,\n  `AIzaSyBcE-OOIbhjyR83gm4r2MFCu4MJmprNXsw`,\n  `AIzaSyB8qGxt4ec15vitgn44duC5ucxaOi4FmqE`,\n  `AIzaSyA8vmApnrHNFE0bApF4hoZ11srVL_n0nvY`\n])",
+		ValidateExpr: `let r = http.get("https://generativelanguage.googleapis.com/v1beta/models?key=" + finding["secret"], {}); r.status == 200 && (r.body contains '"models"') ? {
```
rgmz: Do we *only* want to validate against the Gemini API? @bplaxco had some cool scripts for more specificity about what the key can access -- @bedros-p also has a really neat tool.

jl-aikido: AFAIK the primary risk with a Google API key comes from when it can access Gemini (b/c that grants access to uploaded files, cached contexts, and the ability to rack up a huge bill quickly). I'm sure there are arguments to be made that a Google API key with access to _xyz_ GCP API carries _some_ risk, but we'd need to weigh sending a boat load more requests to ascertain that vs. the actual risk accessing those APIs presents. wdyt?

cmd/generate/config/rules/gcp.go
```diff
@@ -7,6 +7,60 @@ import (
 	"github.com/betterleaks/betterleaks/regexp"
 )
 
+const gcpAPIKeyValidationExpr = `let k = http.get("https://www.googleapis.com/identitytoolkit/v3/relyingparty/getProjectConfig?key=" + finding["secret"], {}); k.status == 400 ? {
+    "result": "invalid",
+    "reason": "Unauthorized"
```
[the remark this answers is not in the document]
zricethezav: what

cmd/generate/config/rules/gcp.go
rgmz: I see it as two distinct questions: (1) is this key live, (2) what risk does it posses? If we *only* want to check Gemini because it's the highest demonstrable risk that's fines; however, I can see a user being confused by a live non-Gemini key being missed.

jl-aikido: GCP's policy AFAIK is that these keys are safe to expose publicly (with the exception of accessing Gemini). I interpret that as meaning unless it can access Gemini, it's not a secret. It's just a publishable key. In those cases, I don't see a need to surface the key in the results, since it's not a secret.

zricethezav: So are we in agreement with the conditions?

rgmz: Yup!

zricethezav:
gonna hit the green button. might be a couple kinks but we can work it out

cmd/generate/config/rules/gcp.go
bplaxco: > GCP's policy AFAIK is that these keys are safe to expose publicly (with the exception of accessing Gemini).

I think Firebase dbs might be another thing these can access that's of any potential risk. I'm not sure about other things. 

And here's a script to get the Google project ID from an API key:

https://github.com/leaktk/hack/blob/main/google-api-key-project-id

bedros-p: Yow, missed this, thanks for the mention! This is a bit long, sorry, but it all just says that despite Google claiming API keys are safe to include in client code, you can still use them for some pretty sick recon :) I definitely think the existence of GCP API keys is worth surfacing regardless of the Gemini API having access or not

When it comes to risks of including an API key in a public client, there's more information leakage than people realize. None of which are particularly considered sensitive but still actionable for recon and good for post-exploitation. I'm still in the process of fully mapping out all the risks but what I have so far:

- Project number as mentioned by @bplaxco but also something extra (mentioned below*)
- You can retrieve oauth consent screen details along with the Oauth consent screen contact information. On sites with more than 2 API keys, this usually means one is a direct developer email. Very good for outreach on responsible disclosure instead of stumbling around trying to contact the company.
- Bruteforce lets you to see what services are enabled on the GCP project it's bound to by hitting discovery docs, good for recon
- You can even view the services enabled on the project that the key *can't* reach, by hitting predictable service account IDs for each service

Currently, all of those are implemented [here, in Fireblazer](https://github.com/bedros-p/fireblazer) under the `-blaze` flag. Without it, it's just the bruteforce discovery docs method. ~430ish discovery docs supported and ~300ish per-project per-service (P4SA) service accounts too, finishes in a matter of a few seconds. There are more service accounts in gcloud but the ones not included are ones that just indicate project enablement on something already caught by the included ones (like storage having several P4SAs). This was made *before* the Gemini API key leak article, because API keys were pretty leaky as-is :)

You can also check if the project bound to the key has ListBuckets permissions. For example, proj 901234567890, you can check authenticatedUser ListBuckets permissions thru an authenticated call to https://storage.googleapis.com/storage/v1/b?project=901234567890 - bad configs tend to allow this.

A very new one that has come up very recently, is the autogenerated bucket when you create an app through AI studio and bind it to the project. It's securely set up, but it's quite predictable - https://www.googleapis.com/storage/v1/b/ai-studio-bucket-901234567890-europe-west2  (europe-west2 is an example - one of my buckets were created with that).

So if the deployment is on https://testing-hi-901234567890.europe-west2.run.app/ the build gzip is on https://storage.googleapis.com/download/storage/v1/b/ai-studio-bucket-901234567890-europe-west2/o/services%2Ftesting-hi%2Fversion-1%2Fcompiled%2Fbuild_artifacts.tar.gz?alt=media

AI Studio stuff is not too relevant here, but I'm sharing it just as a way to say, any service that integrates / manages / automates Cloud run is likely also enumerable with these neat tricks - there are _lots_ of predictable naming systems used around Google that rely on project numbers! 

## Additional partially complete methods / in-research for API key extraction

Rn I'm researching a bunch of methods to squeeze out a bit more data from API keys / proj numbers. If any of these line up with something that could benefit Betterleaks, send me a dm on Discord @bedros_p and we can investigate it together - so far I'm investigating:

- Global method to verify key reach to any Google service on the GCP marketplace
- Existence check if a project ID has an unspecified bucket attached
- Listing & key reach testing of all testIamPermissions endpoints across all Google products (can be used for checking if key is bound to a service account)
- Firebase config thru just an API key / proj number
- Comprehensive listing of all cloudrun integrations and predictable formats

There also exist ways to check enablement for some services without needing an API key from that project besides P4SAs. Good for getting around a key restriction issue for GCP enumeration.

## Important note about project IDs
*You can also get project IDs from just the project number. Project IDs being the consumer-selected unique name for it, even though Google tends to use them interchangeably in their API responses they are very different. Proj IDs are not retrievable thru identitytoolkit, but you can make a good guess thanks to the default appspot name including it - generally ez on Firebase keys. BUT, methods to retrieve proj ID from proj number cannot be included in public source code.

Project IDs are more sensitive and can be used for escalating to bucket retrieval, IAP bypasses (appspot domain, targeted bruteforce), it's quite fun. However, Google considers such mapping to be a security issue, and *will* patch it upon finding out. It is not to be included anywhere.

They also do *not* reward instances of project number to project ID mapping as of recently, but it will be accepted and patched. If you find one, save it for an escalation pathway on Google Cloud instead for a better report. Or reach out to me if you need it as one last missing piece of a gVRP bug report and I'd be glad to collab, as I hypothetically, theoretically, cybersecurically, may have one. Or 7. Just until I find a good esc path...

## Additional notes on Fireblazer

If needed, I'd be glad to bring over the bruteforcer magic to betterleaks, we both use the MIT license and I mainly write in Go. I did my best to make it as library-friendly as possible so it can be dropped in w any other Go util, but if necessary I can port it over in a slimmed down way to cut out the unused parts :)

PR #233, opened 2026-07-09
Remove left preamble for a quick perf win

zricethezav:
We remove `[\w.-]{0,50}?` from left of the identifier because that's easy enough to emulate in Go rather than running it in regex. 

cuts a couple ms from a dir scan of gitlab:

median scan times go from 3.5 -> 3.15.

Still needs more testing


internal/exprruntime/bindings_filter.go
```diff
@@ -78,6 +80,24 @@ func containsAny(s string, terms any) bool {
...
 
+func (rt *runtimeBindings) matchesAnyNearMatch(_ any, patterns any, charsBefore, charsAfter int) bool {
+	return matchesAny(rt.nearMatchText(charsBefore, charsAfter), patterns)
+}
+
+func (rt *runtimeBindings) containsAnyNearMatch(_ any, terms any, charsBefore, charsAfter int) bool {
+	return containsAny(rt.nearMatchText(charsBefore, charsAfter), terms)
+}
+
+func (rt *runtimeBindings) nearMatchText(charsBefore, charsAfter int) string {
```
rgmz: If I'm reading this correctly, there is a slight difference because it will take the whole window and not only "identifier" characters?

zricethezav: `nearMatchText` extracts the match + x character before and after the match

rgmz: Aye. The regex matches things that _look like an identifier_ to improve the context (`password` -> `foo.bar.baz.**password**`).

I don't feel strongly about keeping it, just highlighting the consequence of this change.

zricethezav: Yep yep, it's a change in default behavior but configurable

zricethezav:
Gonna mark `{X}NearMatch` bindings as experimental. These will likely get moved to pure expr once we have a better finding model exposed.

PR #234, opened 2026-07-10
fix(exprruntime): data race evaluating a cached Program concurrently

### LINKED ISSUE #224: internal/exprruntime: data race in compiledProgram.evalBindings() when the same cached filter Program is evaluated concurrently
## Summary

`internal/exprruntime.(*compiledProgram).evalBindings()` has a data race when the same compiled `filter`/`prefilter` `Program` is evaluated concurrently by multiple goroutines — which is the normal execution shape for `detect.Detector.Run`/`DetectSource`, since `sources.Files.Fragments()` fans out one goroutine per file via `fatih/semgroup` (default pool size 40, see `detect.NewDetectorContext`), and every file's fragments are matched against the *same* cached `Program` per rule.

Confirmed with `go test -race` / `go build -race` on both **v1.6.0** and **v1.6.1** (latest release as of this report). Not specific to any downstream consumer's config — reproduces with the stock default `betterleaks.toml` rules that ship with the tool, as long as a rule with a `filter`/`entropy` expression matches in 2+ files scanned concurrently.

## Potential root cause

`internal/exprruntime/runtime.go`:

```go
func (prg Program) evalBindings() bindings {
	if prg.bindings != nil {
		b := cloneBindings(prg.bindings)
		if rt, ok := b["__runtime"].(*runtimeBindings); ok {
			rt.tokenizer = prg.tokenizer          // runtime.go:208
			rt.tokenizerProvider = prg.tokenizerProvider // runtime.go:209
		}
		return b
	}
	return bindings{}
}

func cloneBindings(src bindings) bindings {
	dst := make(bindings, len(src))
	for k, v := range src {
		dst[k] = v
	}
	return dst
}
```

`cloneBindings` is a **shallow** copy: the new map `dst` is per-call, but the value stored under `"__runtime"` is a `*runtimeBindings` pointer copied by reference from `prg.bindings["__runtime"]`. That pointer is the *same* `runtimeBindings` instance for every evaluation of a given cached `Program` — it's created once, at compile time, in `filterBindings`/`prefilterBindings` (`baseBindings(&runtimeBindings{...})`) and stored in `prg.bindings`.

`evalBindings()` then mutates `rt.tokenizer` and `rt.tokenizerProvider` **in place** on that shared instance on every call, with no synchronization. When `EvalFilter`/`EvalPrefilter` is invoked for the same rule's compiled `Program` from multiple goroutines at once (exactly what happens when `sources.Files.Fragments()` scans several files concurrently and each file's fragments are checked against the same rule), two goroutines race on the same two struct fields.

In practice the race is benign in outcome (every writer writes the same value — `prg.tokenizer`/`prg.tokenizerProvider` are invariant for the lifetime of a given compiled `Program`), but it's still undefined behavior under the Go memory model, and it reliably trips `-race`, which makes `-race` unusable for anything that calls into `detect.Detector` with real, multi-file targets.

## Reproduction (self-contained, no downstream project needed)

```bash
# 1. Build betterleaks with the race detector
go install -race github.com/betterleaks/betterleaks@v1.6.1   # also reproduces on v1.6.0

# 2. Generate enough files to force concurrent evaluation of the same
#    filter-bearing default rule (aws-amazon-bedrock-api-key-long-lived,
#    filter: entropy(finding["secret"]) <= 3.0)
mkdir /tmp/bl-race-repro && cd /tmp/bl-race-repro
SECRET="ABSK$(printf 'A%.0s' $(seq 1 109))"
for i in $(seq 1 50); do
  printf 'token := "%s" // %d\n' "$SECRET" "$i" > "f$i.txt"
done

# 3. Scan with the race detector
betterleaks detect --no-git --source . --enable-rule aws-amazon-bedrock-api-key-long-lived --no-banner --redact=0
```

## Observed

```
==================
WARNING: DATA RACE
Write at 0x00c0007c24f8 by goroutine 56:
  github.com/betterleaks/betterleaks/internal/exprruntime.(*compiledProgram).evalBindings()
      .../internal/exprruntime/runtime.go:208 +0x34c
  github.com/betterleaks/betterleaks/internal/exprruntime.(*Runtime).EvalFilter()
      .../internal/exprruntime/runtime.go:192 +0xc8
  github.com/betterleaks/betterleaks/detect.(*Detector).detectFragmentWithRule()
      .../detect/detect.go:979 +0x1540
  github.com/betterleaks/betterleaks/detect.(*Detector).detectFragment()
      .../detect/detect.go:698 +0x798
  ...
  github.com/fatih/semgroup.(*Group).Go.func1()
      .../fatih/semgroup@v1.2.0/semgroup.go:64 +0xc8

Previous write at 0x00c0007c24f8 by goroutine 49:
  [... identical stack, different goroutine ...]
==================
Found 4 data race(s)
```

(4 races per run in this repro: 2 field writes × 2 racing goroutine pairs. Full untrimmed output available on request.)

## Suggested fix

Give each `evalBindings()` call its own private `runtimeBindings` instead of mutating the shared one:

```go
func (prg Program) evalBindings() bindings {
	if prg.bindings != nil {
		b := cloneBindings(prg.bindings)
		if rt, ok := b["__runtime"].(*runtimeBindings); ok {
			rtCopy := *rt
			rtCopy.tokenizer = prg.tokenizer
			rtCopy.tokenizerProvider = prg.tokenizerProvider
			b["__runtime"] = &rtCopy
		}
		return b
	}
	return bindings{}
}
```

This keeps the per-call refresh of `tokenizer`/`tokenizerProvider` (in case that's needed for a lazily-set `Runtime.tokenizerProvider`) without sharing mutable state across concurrent evaluations of the same cached `Program`.

## Environment

- betterleaks v1.6.0 and v1.6.1 (latest)
- go1.26.4 darwin/arm64
- Reproduces with default config, no custom rules required

gtrrz-victor:
## Summary

Fixes #224 — a data race in `internal/exprruntime` when the same cached filter/prefilter `Program` is evaluated concurrently, which is the normal shape for `detect.Detector.Run`/`DetectSource` (one goroutine per file).

## Root cause

A cached filter/prefilter `Program` stores a single shared `*runtimeBindings` (the `__runtime` entry) in its bindings map. `evalBindings` shallow-clones that map but not the pointer, so every concurrent `Eval` mutated shared state:

- `evalBindings` wrote `rt.tokenizer` / `rt.tokenizerProvider` on each call (`runtime.go`, the line reported in #224)
- `failsTokenEfficiency` lazily wrote `rt.tokenizer = rt.tokenizerProvider()`

## Fix

Both values are **program-constant** (the filter cache key already includes the tokenizer pointer), so:

- Bake `tokenizer`/`tokenizerProvider` into the shared bindings **once at compile time** — on the compiling goroutine, before the `Program` is shared — and make `evalBindings` a pure clone with no mutation.
- `failsTokenEfficiency` resolves the tokenizer into a **local** instead of caching it back onto `rt`. No extra cost: the provider (`Detector.Tokenizer`) is already memoized via `sync.Once`.

No behavior change — the reads observe the same program-constant values they did before, just without the concurrent writes.

## Testing

- New `-race` regression test evaluates one cached filter/prefilter `Program` from 50 goroutines (fails on `main`, passes with this change).
- `go test -race ./internal/exprruntime/ ./detect/` — ok (`detect` exercises the real per-file fan-out).
- `go test -race ./...` — clean, no regressions.

### DIFF
--- a/internal/exprruntime/bindings_filter.go
+++ b/internal/exprruntime/bindings_filter.go
@@ -119,16 +119,21 @@ func shannonEntropy(s string) float64 {
 var newlineReplacer = strings.NewReplacer("\n", "", "\r", "")
 
 func (rt *runtimeBindings) failsTokenEfficiency(secret string) bool {
-	if rt.tokenizer == nil {
+	// rt is shared across concurrent evaluations of a cached program, so it must
+	// not be mutated here. Resolve the tokenizer into a local; the provider
+	// (Detector.Tokenizer) is already memoized via sync.Once, so re-resolving is
+	// cheap and there is no need to cache it back onto rt.
+	tke := rt.tokenizer
+	if tke == nil {
 		if rt.tokenizerProvider == nil {
 			return false
 		}
-		rt.tokenizer = rt.tokenizerProvider()
-		if rt.tokenizer == nil {
+		tke = rt.tokenizerProvider()
+		if tke == nil {
 			return false
 		}
 	}
-	return failsTokenEfficiency(rt.tokenizer, secret)
+	return failsTokenEfficiency(tke, secret)
 }
 
 func failsTokenEfficiency(tke *tiktoken.Tiktoken, secret string) bool {
--- a/internal/exprruntime/runtime.go
+++ b/internal/exprruntime/runtime.go
@@ -154,6 +154,15 @@ func (e *Runtime) compile(mode compileMode, expression string, tokenizer *tiktok
 		tokenizerProvider: e.tokenizerProvider,
 		bindings:          programBindings(mode, b),
 	}
+	// Filter/prefilter programs are cached and evaluated concurrently, so their
+	// bindings (including the shared *runtimeBindings) must be immutable after
+	// compile. The tokenizer and provider are program-constant, so bake them in
+	// here — once, on the compiling goroutine, before the program is shared —
+	// rather than writing them on every Eval (which raced). See evalBindings.
+	if rt, ok := prg.bindings["__runtime"].(*runtimeBindings); ok {
+		rt.tokenizer = prg.tokenizer
+		rt.tokenizerProvider = prg.tokenizerProvider
+	}
 
 	e.mu.Lock()
 	e.cache[cacheKey] = prg
@@ -206,14 +215,14 @@ func (e *Runtime) EvalPrefilter(prg Program, attributes map[string]string) (bool
 	return runBool(prg, b, "prefilter")
 }
 
+// evalBindings returns a per-eval shallow copy of the program's bindings. The
+// map is cloned so callers can layer dynamic values (finding, attributes)
+// without racing, while the shared function bindings — including the immutable
+// *runtimeBindings whose tokenizer/provider were fixed at compile time — are
+// safe to share by pointer across concurrent evaluations.
 func (prg Program) evalBindings() bindings {
 	if prg.bindings != nil {
-		b := cloneBindings(prg.bindings)
-		if rt, ok := b["__runtime"].(*runtimeBindings); ok {
-			rt.tokenizer = prg.tokenizer
-			rt.tokenizerProvider = prg.tokenizerProvider
-		}
-		return b
+		return cloneBindings(prg.bindings)
 	}
 	return bindings{}
 }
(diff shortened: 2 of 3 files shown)


gtrrz-victor:
Superseded by #231 ("fix data race w/ runtime bindings"), which landed on `main` first and fixes the same race in `internal/exprruntime` — verified it passes the concurrent-eval regression test under `-race` and scans clean. I was a bit too late here. Closing.

PR #241, opened 2026-07-13
perf(detect): limit regex searches to inferred keyword regions

zricethezav:
This pull request introduces a new mechanism for inferring and restricting the scan region for certain secret-detection rules, improving performance and accuracy by limiting the search area to a region around detected keywords when possible. It also adjusts the generic credential regex to better cap match lengths, and adds tests to validate these behaviors.

### DIFF
--- a/cmd/generate/config/rules/generic.go
+++ b/cmd/generate/config/rules/generic.go
@@ -22,7 +22,7 @@ func GenericCredential() *config.Rule {
 			"passw(?:or)?d",
 			"secret",
 			"token",
-		}, `[\w.=-]{10,150}|[a-z0-9][a-z0-9+/]{11,}={0,3}`, true),
+		}, `[\w.=-]{10,150}|[a-z0-9][a-z0-9+/]{11,149}={0,3}`, true),
 		Keywords: []string{
 			"access",
 			"api",
--- a/detect/search_scope.go
+++ b/detect/search_scope.go
@@ -1,0 +1,270 @@
+package detect
+
+import (
+	"math"
+	"regexp/syntax"
+	"strings"
+	"unicode"
+	"unicode/utf8"
+
+	"github.com/betterleaks/betterleaks/config"
+)
+
+const (
+	maxExactVariants = 4096
+	maxExactBytes    = 64
+)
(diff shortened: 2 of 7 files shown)


cmd/generate/config/rules/generic.go
```diff
@@ -22,7 +22,7 @@ func GenericCredential() *config.Rule {
 			"passw(?:or)?d",
 			"secret",
 			"token",
-		}, `[\w.=-]{10,150}|[a-z0-9][a-z0-9+/]{11,}={0,3}`, true),
+		}, `[\w.=-]{10,150}|[a-z0-9][a-z0-9+/]{11,149}={0,3}`, true),
```
rgmz, line 25: What's the significance of 149?

detect/search_scope_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/betterleaks/betterleaks/regexp"
+)
+
+func TestInferKeywordScope(t *testing.T) {
+	for _, test := range []struct {
+		name     string
+		pattern  string
+		keywords []string
+		want     int
+	}{
```
rgmz, line 20: What about keywords _after_ the match?

detect/search_scope.go
```diff
@@ -0,0 +1,270 @@
...
+)
+
+const (
+	maxExactVariants = 4096
+	maxExactBytes    = 64
+)
+
+// inferKeywordScope returns a conservative byte radius around keyword matches.
+// Zero keeps full-fragment evaluation.
+func inferKeywordScope(rule config.Rule) int {
```
rgmz, line 20: Implicit match spans was a persistent source of bugs for TruffleHog. IMO this shouldn't be enabled by default without data and testing.

cmd/generate/config/rules/generic.go
zricethezav: needed _some_ value, it's 100% arbitrary

rgmz: This will likely result in missed secrets. Would you be opposed for something longer -- maybe closer to 500?

PR #250, opened 2026-07-16
docs: add Go installation instructions to README

RoseSecurity:
## what and why

- Add additional Go install method to documentation
- Provide other ways to install betterleaks


rgmz (approved):
Thanks!

PR #253, opened 2026-07-20
Rules/jul 17 batch 1

zricethezav:


cmd/generate/config/rules/ebay.go
```diff
@@ -0,0 +1,68 @@
...
+		Regex: utils.GenerateSemiGenericRegex(
+			[]string{`ebay(?:[_. -]*(?:client|api))?[_. -]*(?:secret|key)`},
+			`PRD-[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4,12}`,
+			false,
+		),
+		Keywords: []string{"ebay"},
+		RequiredRules: []*config.Required{
+			{RuleID: "ebay-client-id"},
+		},
+		ValidateExpr: `let clientID = captures["ebay-client-id"];
```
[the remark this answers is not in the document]
zricethezav: false, all capture maps get loaded up with required rules

cmd/generate/config/rules/fullstory.go
```diff
@@ -0,0 +1,40 @@
...
+
+func FullStoryAPIKey() *config.Rule {
+	// define rule
+	r := config.Rule{
+		RuleID:      "fullstory-api-key",
+		Description: "FullStory API key.",
+		Regex: utils.GenerateSemiGenericRegex(
+			[]string{`(?:fullstory|fs_api|fullstory_api)`},
+			`(?:na1|eu1)\.[A-Za-z0-9]{20,}`,
+			true,
```
rgmz, line 17: RE: #241, for stuff like this would we be able to do something like this, and have an implicit trust that a prefix isn't necessary? (because the keyword is in the chunk/localized span)
```go
Regex: regexp.MustCompile(`(?:na1|eu1)\.[A-Za-z0-9]{20,}`),
Keywords: []string{"fullstory", "fs_api"},
```

cmd/generate/config/rules/fullstory.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/betterleaks/betterleaks/config"
+)
+
+func FullStoryAPIKey() *config.Rule {
+	// define rule
+	r := config.Rule{
+		RuleID:      "fullstory-api-key",
+		Description: "FullStory API key.",
+		Regex: utils.GenerateSemiGenericRegex(
+			[]string{`(?:fullstory|fs_api|fullstory_api)`},
```
rgmz, line 15: Also:
```suggestion
			[]string{`(?:fullstory|fs_api)`},
```

zricethezav: A benefit of having both is that `_api` eats less into the characters matched after the identifier and before the secret or assignment group. Pros and cons for both I suppose

PR #254, opened 2026-07-21
chore(container): add dockerignore files; update golang version

pstoeckle:
- use `.dockerignore` files to keep build context smaller
- update go version in `Dockerfile` to version in `go.mod`


Dockerfile.dockerignore
```diff
@@ -0,0 +1,16 @@
...
+!go.mod
+!go.sum
+!internal
+!logging
+!regexp
+!report
+!report_templates
+!sources
+!version
+!main.go
```
rgmz, line 15: Thanks for the PR, @pstoeckle! I believe the Go files are intentionally included because it's a multi-stage image; the final result should only contain the binary.
https://github.com/betterleaks/betterleaks/blob/b49c53582daddd378671a636cb5e9cc70bef5c5b/Dockerfile#L1-L5

@zricethezav can confirm.

pstoeckle: > I believe the Go files are intentionally included

Not sure if I understand this correctly, but 

the

```
*
```

in the `.dockerignore` states that **ALL** files should be ignored.
And the following

```
!cmd
!config
```

i.e., the lines with the leading `!` state that these files/folders should **NOT** be ignored and, thus, be included/copied in the `COPY . . ` statement in the `Dockerfile`.

PR #259, opened 2026-07-21
Fix PostHog key length ranges and add personal key validator

feliperalmeida:
## Summary

PostHog rules have a static size that miss some keys. Key length has changed over time, so I'm updating those rules to cover all PostHog key sizes.

Also adding a **personal** key validator, validating against US and EU production instances. 

PostHog **project** API keys are meant to be public, so not adding a validator for those - just editing the description to be more accurate on the key purpose.

### DIFF
--- a/cmd/generate/config/rules/posthog.go
+++ b/cmd/generate/config/rules/posthog.go
@@ -50,8 +50,10 @@ func PostHogPersonalAPIKey() *config.Rule {
 		Keywords: []string{"phx_"},
 		Filter:   `entropy(finding["secret"]) <= 3.0`,
 		// A valid key hits us/eu.posthog.com and returns 200 (has user:read) or
-		// 403 (authenticated but missing the user:read scope); an unknown/revoked
-		// key returns 401. Keys are region-bound, so fall back from US to EU.
+		// 403 (authenticated but missing the user:read scope); a revoked/unknown
+		// key returns 401. Keys are region-bound, so any non-valid US response
+		// (401, or a 429/5xx outage) falls back to EU. Only report "invalid" when
+		// both regions definitively return 401 — if either errored, stay unknown.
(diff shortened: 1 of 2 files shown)


cmd/generate/config/rules/posthog.go
```diff
@@ -9,16 +9,28 @@ import (
...
 		RuleID:      "posthog-project-api-key",
-		Description: "Detected a PostHog Project API Key, which may expose product analytics data and event tracking to unauthorized access.",
-		Regex:       utils.GenerateUniqueTokenRegex(`phc_[a-zA-Z0-9_\-]{43}`, true),
-		Keywords:    []string{"phc_"},
-		Filter:      `entropy(finding["secret"]) <= 3.0`,
+		Description: "Detected a PostHog Project API Key, a public write-only token used to send events to a PostHog project.",
+		// "phc_" + a random token. The encoding has changed over time, so the body
+		// length varies across keys still in the wild:
+		//   - base62(32 bytes)          → 41-43 chars (2021 .. early 2026)
+		//   - base57(32 bytes), top-bit → exactly 44 chars (current, since #52495)
```
rgmz, line 16: Thanks, @feliperalmeida !

Out of curiosity, is the key character set document anywhere that we can link to?

feliperalmeida: I'm afraid it's not documented anywhere other than [in code](https://github.com/PostHog/posthog/blob/951fb99a1c9c059dae89d7ed14f98e9c4d3a6e8b/posthog/models/utils.py#L146) and git history.

zricethezav (approved):

zricethezav:
@feliperalmeida thank you so much!

zricethezav:
@feliperalmeida 
> PostHog project API keys are meant to be public, so not adding a validator for those - just editing the description to be more accurate on the key purpose.

If they're meant to be public we might remove them by default. Recently we started removing all the public/publishable API keys

PR #260, opened 2026-07-21
More rules and rule updates (p-r)

zricethezav:


cmd/generate/config/rules/pinecone.go
```diff
@@ -0,0 +1,69 @@
...
+		`PINECONE_API_KEY=pcsk_short`,
+		`PINECONE_API_KEY=pcsk_abc12_xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx`,
+	}
+	return utils.Validate(r, tps, fps)
+}
+
+func PineconeLegacyAPIKey() *config.Rule {
+	// define rule
+	r := config.Rule{
+		RuleID:      "pinecone-legacy-api-key",
```
rgmz: Aside: we'd benefit from a versioning scheme a-la NoseyParker (`pinecone.apikey.1`) or TruffleHog (`pinecone-api-key:1`).

Should I make an issue for this?

cmd/generate/config/rules/pinterest.go
```diff
@@ -0,0 +1,40 @@
...
+  } : validate.unknown(r)`,
+		Filter: utils.MinEntropy(3.5),
+	}
+
+	// validate
+	tps := []string{
+		"PINTEREST_ACCESS_TOKEN=pina_" + secrets.NewSecretWithEntropy(`[A-Za-z0-9_-]{40}`, 3.5),
+	}
+	fps := []string{
+		`PINTEREST_ACCESS_TOKEN=pina_short`,
```
rgmz, line 36: An interesting one worth testing. This pattern seems prone to false-positives on GitHub search (albeit, not filtered with entropy).
```suggestion
	fps := []string{
	    `#: model:product.template,name:point_of_sale.pina_colada_product_template`,
		`PINTEREST_ACCESS_TOKEN=pina_short`,
```

cmd/generate/config/rules/pinecone.go
zricethezav: if legacy versions are still accepted, yes. otherwise no need to keep old rules around that cant validate against anything

zricethezav: updated the pinecone rules for this PR. We should tackle all the other rules in another issue.

PR #267, opened 2026-07-26
feat: secret revocation via rule-level 'revoke' expressions

### LINKED ISSUE #212: Feature request: Include command to revoke a secret (where supported)
Including a command to revoke a secret gives engineers to easy option to revoke their secret programmatically (where supported, e.g. [Buildkite](https://buildkite.com/docs/apis/rest-api/access-token#revoke-the-current-token). The command could be displayed along with scan data, and the revoke only occurs when requested.

omarbabba779xx:
Closes #212.

Adds a `revoke` Expr expression alongside the existing `validate` one, letting a rule deactivate a live secret at its provider (e.g. deleting a Buildkite access token via its `DELETE /v2/access-token` endpoint).

## Design

Revocation is deliberately **never** run automatically during a scan - it's a destructive, irreversible action against a real credential. It only runs via a new `betterleaks revoke <report-file>` command, against findings already in a report, and only for rules whose config defines a `revoke` expression. Findings for rules with no `revoke` expression are reported as unsupported and skipped.

By default each revocation is confirmed interactively (`Revoke this secret now? This cannot be undone. [y/N]`). Flags:
- `--fingerprint` - only consider one specific finding
- `--dry-run` - preview without calling any provider
- `--yes` - skip the confirmation prompt (e.g. CI)
- `--continue-on-error` - keep going after a failure instead of stopping
- `--output` - write an updated report with revoked findings marked

`revoke` uses the same result-map format as `validate` (`{"result": "revoked", ...}`); anything other than `"revoked"` is treated as a failed revocation, never silently reported as success.

## What's included

- `config`: `revoke` TOML field, `[extend]` inheritance support, `Config.CompileRevocation()`
- `internal/exprruntime`: `revoke` namespace binding (`revoke.unknown()`), plus a new `http.delete` binding since revocation is most commonly a DELETE call
- `cmd/revoke.go`: the CLI command, with the core evaluation logic (`revokeOne`) kept separate from I/O so it's directly unit-testable
- `cmd/config.go`: wired `revoke` expression compilation into `betterleaks config check`/`config show`'s eager validation, matching how `validate` is already checked there - otherwise a broken `revoke` expression would pass `config check` clean and only surface mid-revocation against a real credential
- A real proof-of-concept `revoke` expression on the existing `buildkite-user-access-token` rule, using the same DELETE endpoint referenced in #212
- README and `docs/config.md` updated with a Revocation section

## Testing

New tests at every layer: config parsing/inheritance/compilation (`config/revoke_test.go`), the `revoke` Expr binding end-to-end against a mock HTTP server (`internal/exprruntime/bindings_revoke_test.go`), `revokeOne`/`loadFindingsReport`/formatting helpers (`cmd/revoke_test.go`), and the config-check fix (`cmd/config_test.go`). Full existing suite passes with zero regressions (`go build ./...`, `go vet ./...`, `go test ./...`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ Development is supported by
 | :--- | :--- |
 | **Expr-based filtering** | Write contextual rule filters that evaluate fragment (data chunks) attributes (like git author, commit message, and file path) and finding data to reduce false positives. If you're coming from Gitleaks, think of this feature as a more expressive `[[allowlist]]` system. |
 | **Secrets Validation** | Validate if a detected secret is active by making asynchronous HTTP requests directly from within the rule definition using Expr. |
+| **Secrets Revocation** | Deliberately revoke a live secret at its provider via `betterleaks revoke`, using a rule-level `revoke` expression. Manual and explicit only — never run during a scan. |
 | **Token Efficiency filtering** | Filter out natural language false positives by using BPE tokenization to measure how "rare" or non-human a string is. |
 | **Fast scans** | Achieve fast performance through sane default parallelization settings, ahocorasick keyword filters, and re2. |
 | **New Sources** | Support for sources like GitHub, GitLab, Hugging Face, S3, and more. It's easy to add new sources too!   |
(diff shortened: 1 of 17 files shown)


omarbabba779xx:
Closing in favor of a fresh PR with the review fixes squashed in cleanly from the start.

PR #277, opened 2026-07-30
validate cmd

zricethezav:


docs/scanning.md
```diff
@@ -528,6 +529,117 @@ Objects in `GLACIER`, `GLACIER_IR`, and `DEEP_ARCHIVE` storage classes are skipp
...
+  secret: env.AWS_ACCESS_KEY_ID,
+  components: {
+    "aws-secret-access-key": env.AWS_SECRET_ACCESS_KEY
+  }
+}' | betterleaks validate \
+	--rule-id aws-access-token
+```
+
+The JSON fields are `secret` (required), `components`, `captures`, and
+`attributes`; the latter three are string-to-string objects. For less sensitive
```
rgmz: Do we have examples of the latter two in practice? Captures at least feels like we should do a better job for the user experience of "components".

docs/scanning.md
```diff
@@ -528,6 +529,117 @@ Objects in `GLACIER`, `GLACIER_IR`, and `DEEP_ARCHIVE` storage classes are skipp
...
+interactive use, supply each rule declared in `[[rules.required]]` as
+`--component rule-id=secret`. Use `--capture name=value` when a validation
+expression needs a named regex capture that cannot be reconstructed from the
+credential. A component capture uses `--capture rule-id:name=value`.
+Source-aware custom validators can receive attributes through repeated
+`--set-attr key=value` flags.
```
rgmz: I've been living under a rock for a few months. Are there any validators that do this today?

zricethezav: negative

PR #281, opened 2026-08-05
feat(rule): validation for DockerHub tokens

rgmz:
This adds a basic validator for Docker Hub `pat_` and `oat_` tokens, including a best-effort "username" rule as that's required for validation.

**I would not merge this yet**. As discussed, the current implementation requires `dockerhub-username` to also be found; however, `dckr_pat_` and `dckr_oat_` are high confidence patterns which should be reported even without a username.
image

-------

**Edit:** depends on #282


zricethezav:
@rgmz merged #282 btw

PR #282, opened 2026-08-05
introduce components field, start deprecating rules.required

zricethezav:
(codex gpt-5.6-sol xhigh implemented)

Consolidates multipart rule configuration under a new `components` field. Components are required by default, or can be marked with `optional = true`. (default is `optional=false` which implies "required component"). This is technically a breaking change because findings now expose `ComponentSets` instead of `RequiredSets`, but I imagine most folks aren’t consuming that field directly (could be wrong). Old `[[rules.required]]` config syntax is still accepted for backwards compatibility, while generated and displayed configs use the new format.


config/config.go
```diff
@@ -63,11 +63,19 @@ type rawRule struct {
...
-	SkipReport      bool                `toml:"skipReport"`
-	TokenEfficiency bool                `toml:"tokenEfficiency"`
+	Allowlists []*rawRuleAllowlist `toml:"allowlists"`
+
+	// Components is a pointer so config extension can distinguish omission
+	// from an explicit empty list.
+	Components *[]*rawComponent `toml:"components"`
+
+	// Deprecated: translated to required components when Components is absent.
+	// TODO: Remove this in 9.x.
```
zricethezav: will drop comment

config/config.go
```diff
@@ -244,17 +257,35 @@ func (rc *rawConfig) translate(depth int) (*Config, error) {
...
-		for _, r := range vr.Required {
-			if r.ID == "" {
-				return nil, fmt.Errorf("%s: [[rules.required]] rule ID is empty", cr.RuleID)
+		if vr.Components != nil {
+			cr.componentsSet = true
+			for _, component := range *vr.Components {
+				if component == nil {
+					component = &rawComponent{}
+				}
+				cr.Components = append(cr.Components, &Component{
```
rgmz, line 266: Should this be `map[int]*rawComponent`? I had the dumb thought: what if you did this:
```toml
components = [
    { id = "aikido-client-id" },
    { id = "aikido-client-id" },
]
```

detect/match_context.go
```diff
@@ -1,110 +1,26 @@
...
-		spec.LinesBefore = max(l.before, l.bidirected)
-		spec.LinesAfter = max(l.after, l.bidirected)
-	} else if hasC {
-		spec.Mode = ContextModeCols
-	} else {
-		return MatchContextSpec{}, fmt.Errorf("invalid match-context spec %q", s)
-	}
-
-	return spec, nil
+	return contextwindow.Parse(s)
```
rgmz: Is this to preserve backwards compatibility?

internal/contextwindow/contextwindow.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+
+		directionMarker := matches[1]
+		amount, _ := strconv.Atoi(matches[2])
+		unit := strings.ToUpper(matches[3])
+		if unit == "" {
+			unit = "C"
+		}
+
+		if unit == "L" {
```
rgmz: Should this loudly fail if `unit` isn't C or L?

zricethezav: yea me thinks so

config/config.go
zricethezav: fails if duplicates detected now

detect/match_context.go
zricethezav: it was, but really contextwindow should be internal only. good catch

PR #288, opened 2026-08-10
additional gitlab incoming mail token detection

jl-aikido:
The existing GitLab incoming mail token detector misses two cases:
(1) Users setting custom token prefixes. 
(2) Tokens found in email addresses before the `glimt-` prefix existed. Easily found a dozen cases of these online. There are probably many more. 

This PR creates a new rule `GitlabIncomingMailAddressToken` that provides detection for these two cases. Since this will also sweep up existing `glimt-` tokens surfaced by the `GitlabIncomingMailToken` rule, it adds a specificity value, so that the `GitlabIncomingMailToken` result takes priority and we avoid duplicates.


zricethezav (approved):
love it

PR #302, opened 2026-08-17
fix: resolve Sema concurrency bug and cleanup rule TODOs

darubaidnazir:
This commit fixes a critical concurrency bug in detect.go where the semgroup's Wait() method was never called. This caused errors from background scanning goroutines to be silently swallowed.

Additionally, this commit resolves several TODOs in the rules package:
- Implement validation rules for dropbox short/long lived tokens.
- Uncomment and fix true/false positive test cases in discord, flyio, grafana, huggingface, infracost, jwt, telegram, and generic rules.


zricethezav:
I have no problem with AI PRs but mixing TODOs with a critical bug is not acceptable

PR #308, opened 2026-08-18
Clone strings to allow release of fragment.Raw 

### LINKED ISSUE #307: memory use increases with findings, can run out of memory
betterleaks used 12 GB of memory while scanning a large data set which had thousands of findings. My system ran out of memory and killed betterleaks before it completed the scan. The root of the problem is that `secret` and some other strings in a finding are substrings of `fragment.Raw` and prevent the fragment from being garbage collected. This chart shows how the memory grows with findings. (I stopped this test around 170 MB. This isn't the run that used 12 GB.)

Image

gpflaum:
Fixes #307

Substrings of `fragment.Raw` in findings prevented gc of the `Raw` buffer, causing high memory use when there are many findings. Clone strings to allow release of `fragment.Raw`.

Image

### DIFF
--- a/detect/detect.go
+++ b/detect/detect.go
@@ -887,7 +887,9 @@ func (d *Detector) detectFragmentWithRule(fragment sources.Fragment,
 	// Reuse the matches slice from above instead of calling FindAllStringIndex again.
 	for _, matchIndex := range matches {
 		// Extract secret from match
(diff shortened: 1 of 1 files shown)


zricethezav:
oooof. Good catch. Probably need to create some CI check that measures anomalies in mem usage and scan times

zricethezav:
thanks for being on top of it @gpflaum!

PR #313, opened 2026-08-20
Clone box context so fragment.Raw can be released

gpflaum:
Follow-up to #308. I missed `finding.exprContext`.


zricethezav:
😬 thanks again @gpflaum

PR #322, opened 2026-08-24
fix: make sumoLogic accessID prefix case-sensitive

### LINKED ISSUE #321: update `sumoLogic` regex
The `SumoLogicAccessID` rule currently uses `isCaseInsensitive = false` to avoid matching uppercase `SU` prefixes.

This updates the flag to `true` and updates the regex to use `(?-i:su)`.

Useles5:
fixes #321 

This updates the flag to `true` and updates the regex to use `(?-i:su)`.


rgmz (approved):
Thanks!

PR #326, opened 2026-08-25
rm more deprecated fields

zricethezav:


detect/detect.go
```diff
@@ -162,66 +164,6 @@ type Detector struct {
...
-	// effect. Consume Run results and retain them in the caller when needed.
-	SkipFindingAppend bool
-
-	// ----------------------------------------------------------------
-	// DEPRECATED fields below, to be removed in the next major version
-	//
-	//
-	// report-related settings.
-	// Deprecated: detect should not handle reporting
-	ReportPath string
```
rgmz, line 177: @bplaxco Does LeakTK use any of these?

zricethezav: It's time

PR #327, opened 2026-08-26
fix(config): preserve targeted allowlists across extensions

### LINKED ISSUE #300: Global allowlist with targetRules is silently dropped when the config is extended
## Description

A global allowlist scoped with `targetRules` works as expected when it is declared in the outermost config, but is **silently discarded when that same config is loaded as an extension** of another config (via `[extend] path` or `[extend] useDefault`).

There is no error and no log output, at any log level including `trace`. The config parses successfully, the target rule still runs, and the allowlist simply has no effect.

Unscoped global allowlists and rule-level `[[rules.allowlists]]` are both carried through `extend` correctly. Only the `targetRules` form is lost.

## Reproduce

Three files in an empty directory:

`secret.md`

```
# Example page

    key = 7Rq2XkPzL9mVt4WbN8sJyH3fD6gC1aQe
```

`parent.toml`

```toml
[extend]
useDefault = true

[[allowlists]]
description = "documentation example, not a real credential"
targetRules = ["generic-api-key"]
regexes = ['''7Rq2XkPzL9mVt4WbN8sJyH3fD6gC1aQe''']
```

`child.toml`

```toml
[extend]
path = 'parent.toml'
```

Then:

```console
$ betterleaks dir . -c parent.toml --no-banner
no leaks found

$ betterleaks dir . -c child.toml --no-banner
leaks found: 1
```

The reported finding is `secret.md:3` / `generic-api-key` (confirmed with `--report-format json`), i.e. exactly the occurrence the allowlist is meant to cover.

`child.toml` adds nothing except extending `parent.toml`, so both runs should agree.

**Expected:** `no leaks found` in both cases — the allowlist in `parent.toml` applies either way.

**Actual:** the allowlist is ignored as soon as `parent.toml` is the extended config rather than the outermost one.

## Which allowlist forms are affected

Same secret, same extending child, varying only the shape of the allowlist in the parent:

| Allowlist form in the parent config | Parent used directly | Parent extended by a child |
| --- | --- | --- |
| Global `[[allowlists]]` with `targetRules` | no leaks | **leak reported** |
| Global `[[allowlists]]` without `targetRules` | no leaks | no leaks |
| Rule-level `[[rules.allowlists]]` | no leaks | no leaks |

`targetRules` declared in the *outermost* config also works correctly, so the feature itself is fine — only its propagation through `extend` is broken. This is not related to `maxExtendDepth`; the rules themselves are all present and active, they just never receive the allowlist.

## Root cause

> **Please read the [Disclosure](#disclosure) section below before weighing this part.** I reproduced the behaviour above myself, but the code analysis in this section and the fix suggestion that follows are not mine and I cannot personally vouch for them.

All in `config/config.go` (line numbers from `706dbe29`).

At parse time, an allowlist with `targetRules` is deliberately kept out of the global list and parked in `ruleAllowlists`, a map local to the `translate()` call (declared line 189):

```go
// line 336
// Allowlists with |targetRules| aren't added to the global list.
if len(a.TargetRules) > 0 {
    for _, ruleID := range a.TargetRules {
        // It's not possible to validate |ruleID| until after extend.
        ruleAllowlists[ruleID] = append(ruleAllowlists[ruleID], allowlist)
    }
} else {
    c.Allowlists = append(c.Allowlists, allowlist)
}
```

That map is only drained onto the rules under a `depth == 0` guard (line 364), after extends have resolved:

```go
// line 376
// Populate targeted configs.
for ruleID, allowlists := range ruleAllowlists {
    rule, ok := c.Rules[ruleID]
    if !ok {
        return nil, fmt.Errorf("[[allowlists]] target rule ID '%s' does not exist", ruleID)
    }
    rule.Allowlists = append(rule.Allowlists, allowlists...)
    c.Rules[ruleID] = rule
}
```

`extendPath()` and `extendDefault()` both call `translate(depth + 1)`. For an extended config `depth > 0`, so the block never runs and `ruleAllowlists` is discarded when `translate()` returns. Because such allowlists were never appended to `c.Allowlists` either, nothing else carries them.

This also explains why the other two forms survive: unscoped global allowlists are merged by `extend()` (line 719, `c.Allowlists = append(c.Allowlists, extensionConfig.Allowlists...)`), and rule-level allowlists travel inside the `Rule` structs handled by the rule-merge loop.

## Secondary effect: validation is skipped for extended configs

The same `depth == 0` guard also skips `rule.Validate()`, the component rule-ID existence check, and the `[[allowlists]] target rule ID '%s' does not exist` error (line 380) for extended configs.

So a **typo'd** rule ID in an extended config's `targetRules` is silently ignored rather than reported — the opposite of what the comment at line 338 intends ("It's not possible to validate `ruleID` until after extend"). A user who misspells a rule ID in an extended config gets no feedback at all.

## Suggested fix

Simply moving the population block out of the `depth == 0` guard would not be correct: at `depth > 0` the target rule may legitimately not exist yet, since it can be defined by the config that is doing the extending.

A fix that preserves the current post-extend validation semantics would be to carry the pending scoped allowlists on the `Config` struct, merge them in `extend()` alongside the existing `c.Allowlists` merge at line 719, and drain the accumulated map once at `depth == 0`. That resolves both the dropped allowlist and the skipped validation.

`config/config_test.go` already has a table-driven set of extend cases with fixtures in `testdata/config/valid/` (`extend_rule_allowlist_base.toml`, `extend_rule_override_*`), so a regression test fits the existing pattern.

I would rather leave the implementation to you than send a PR I am not in a position to review myself (see Disclosure). What I can offer is testing: I am happy to run any candidate fix against both the reproducer above and our real repository and report back.

## How this is hit in practice

This is not a contrived setup. MegaLinter v10 forwards its `EXCLUDED_DIRECTORIES` to betterleaks by generating a wrapper config that pulls the repository's own `.betterleaks.toml` in via `[extend] path`:

```toml
[extend]
path = '/tmp/lint/.betterleaks.toml'

[allowlist]
paths = ['\.git/', 'megalinter\-reports/']
```

Any repository whose `.betterleaks.toml` uses `targetRules` therefore loses that allowlist the moment it is run through MegaLinter v10, with no indication why. In our case a documentation repository's two allowlisted example secrets started failing CI after the upgrade.

The workaround was to drop `targetRules` and accept a broader, unscoped allowlist: https://github.com/osism/osism.github.io/pull/1071 — that PR has the full investigation, including the matrix above and the point at which we narrowed it down to `targetRules` alone.

Worth noting as a general consequence: the natural workarounds for this bug — unscoping the allowlist or disabling the rule — are both broader than what the user actually intended to allow.

## Impact

To be clear about severity: this cannot hide a real secret, since allowlists only suppress findings. The failure mode is false positives, not false negatives. The cost is the silence — a correct-looking config with no effect and no diagnostic is expensive to track down.

## Disclosure

To be transparent about which parts of this report I can stand behind:

**Reproduced by hand.** I ran the reproducer above myself, and separately confirmed the behaviour end to end in our own repository through MegaLinter. I verified it against the official `ghcr.io/betterleaks/betterleaks:v1.7.3` image as well as the copy bundled in MegaLinter, so the observed behaviour and the comparison table are first-hand.

**Not verified by me: the code analysis.** The root cause explanation, the `config/config.go` references, the reasoning about `depth`, the secondary point about skipped validation, and the suggested fix were all produced with AI assistance (Claude). I am not a Go developer and I am genuinely not able to judge whether that reasoning is correct. It is included because it seemed more useful to you than a bare bug report, but please treat it strictly as a lead to check rather than a conclusion — and disregard it entirely if it is off base. The reproducer stands on its own either way.

I would rather say this plainly up front than have you spend time on analysis I cannot back up.

## Environment

- betterleaks 1.7.3 (as shipped in the MegaLinter v10.0.0 `documentation` flavor image)
- Also present at `main` / `706dbe29`
- Linux, amd64

Possibly inherited rather than new: `git log -S` places this code in `25a0226` ("rename gitleaks -> betterleaks"), so it predates betterleaks and may well affect gitleaks too. I have not verified that against a gitleaks binary.

mameikagou:
Fixes #300.

Preserve targeted global allowlists until path, default, and nested extensions finish assembling, then validate and attach them to the final rules. This prevents extended configs from silently dropping `targetRules` filters.

Validation:
- `go test -race -p 2 ./config`
- `go test -p 2 ./...`
- `go build -p 2 ./...`
- `go vet ./config`

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -735,6 +735,11 @@ func (c *Config) extend(extensionConfig *Config) {
 			c.pendingRuleAllowlists = make(map[string][]*Allowlist)
 		}
 		for ruleID, allowlists := range extensionConfig.pendingRuleAllowlists {
+			if _, disabled := disabledRuleIDs[ruleID]; disabled {
+				if _, inherited := extensionConfig.Rules[ruleID]; inherited {
+					continue
+				}
+			}
 			c.pendingRuleAllowlists[ruleID] = append(c.pendingRuleAllowlists[ruleID], allowlists...)
 		}
 	}
--- a/config/config.go
+++ b/config/config.go
@@ -141,6 +141,11 @@ type Config struct {
 	// Translated from global Allowlists regex/stopword checks.
 	Filter string
 
+	// pendingRuleAllowlists carries targeted global allowlists through config
+	// extensions until all rules have been assembled. It is consumed and
+	// cleared before the translated config is returned to callers.
+	pendingRuleAllowlists map[string][]*Allowlist
+
 	// prefilterProgram and filterProgram hold global programs compiled by
 	// CompileFilters. Per-rule filter and validation compilation is lazy.
 	prefilterProgram exprruntime.Program
@@ -183,10 +188,10 @@ func Default() (*Config, error) {
 
 func (rc *rawConfig) translate(depth int) (*Config, error) {
 	var (
-		keywords       = make(map[string]struct{})
-		orderedRules   []string
-		rulesMap       = make(map[string]Rule)
-		ruleAllowlists = make(map[string][]*Allowlist)
+		keywords              = make(map[string]struct{})
+		orderedRules          []string
+		rulesMap              = make(map[string]Rule)
+		pendingRuleAllowlists = make(map[string][]*Allowlist)
 	)
 
 	// Validate individual rules.
@@ -309,6 +314,7 @@ func (rc *rawConfig) translate(depth int) (*Config, error) {
 		BetterleaksMinVersion: rc.BetterleaksMinVersion,
 		Prefilter:             rc.Prefilter,
 		Filter:                rc.Filter,
+		pendingRuleAllowlists: pendingRuleAllowlists,
 	}
 
 	c.Path = rc.path
@@ -337,7 +343,7 @@ func (rc *rawConfig) translate(depth int) (*Config, error) {
 		if len(a.TargetRules) > 0 {
 			for _, ruleID := range a.TargetRules {
 				// It's not possible to validate |ruleID| until after extend.
-				ruleAllowlists[ruleID] = append(ruleAllowlists[ruleID], allowlist)
+				pendingRuleAllowlists[ruleID] = append(pendingRuleAllowlists[ruleID], allowlist)
 			}
 		} else {
 			c.Allowlists = append(c.Allowlists, allowlist)
@@ -374,7 +380,7 @@ func (rc *rawConfig) translate(depth int) (*Config, error) {
 		}
 
 		// Populate targeted configs.
-		for ruleID, allowlists := range ruleAllowlists {
+		for ruleID, allowlists := range c.pendingRuleAllowlists {
 			rule, ok := c.Rules[ruleID]
 			if !ok {
 				return nil, fmt.Errorf("[[allowlists]] target rule ID '%s' does not exist", ruleID)
@@ -382,6 +388,9 @@ func (rc *rawConfig) translate(depth int) (*Config, error) {
 			rule.Allowlists = append(rule.Allowlists, allowlists...)
 			c.Rules[ruleID] = rule
 		}
+		// Pending allowlists are only needed while assembling the config. Do not
+		// retain extension bookkeeping in the runtime config.
+		c.pendingRuleAllowlists = nil
 
 	}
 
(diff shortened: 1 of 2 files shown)


mameikagou:
Addressed in 116c479: pending allowlists are skipped only when their inherited target rule is explicitly disabled; genuinely missing targets still fail validation. Added regression coverage for both cases.

PR #330, opened 2026-08-26
stream json-like reporters

zricethezav:


zricethezav:
gonna wait for this to be shipped in v2 https://github.com/betterleaks/betterleaks/tree/v2

gpflaum:
@zricethezav Do you have an idea of when the v2 branch will be at a release candidate stage? Would help me decide whether to keep doing some local work with the main branch, or wait a bit and focus on v2. It's fine if you don't want to put any timeframe on it.

zricethezav:
@gpflaum hoping to get a candidate release this month. And actually, maybe we should just get this out the door since most folks use json reporting anyways.

PR #344, opened 2026-09-05
Reuse config across directory scan paths

### LINKED ISSUE #343: `dir` reloads and recompiles the config once per path argument (~16 ms each), so a file list of a few thousand paths takes minutes for a sub-second scan
## Summary

`betterleaks dir` with many file paths as arguments is dominated by per-argument config loading, not by scanning. Passing the same files as a single directory argument scans the same bytes about 100× faster.

Version: betterleaks 1.8.1 (`aqua:betterleaks/betterleaks`), macOS 26 arm64; the same ratio reproduces on ubuntu-latest in GitHub Actions.

## Reproduction

In a git repository with a few thousand tracked files (ours: 5,656 files, 57 MB) and a `.betterleaks.toml` that extends the default config (`[extend] useDefault = true`):

```sh
export BETTERLEAKS_CONFIG=.betterleaks.toml

# whole tree as one argument
betterleaks dir --redact --no-banner .
#   scanned ~59863725 bytes (59.86 MB) in 758ms

# the same tree as explicit file arguments
betterleaks dir --redact --no-banner $(git ls-files)
#   scanned ~57143805 bytes (57.14 MB) in 1m27.3s
```

The cost is linear in the number of path arguments, not in bytes:

| path arguments | bytes scanned | wall time |
|---|---|---|
| 100 | 385 KB | 1.45 s |
| 500 | 4 MB | 8.1 s |
| 5,656 | 57 MB | 87 s |
| `.` (1) | 60 MB | 0.76 s |

Passing the config via `--config`, `BETTERLEAKS_CONFIG`, `BETTERLEAKS_CONFIG_TOML`, or a `.betterleaks.toml` in the cwd makes no difference (each ~8 s for 500 files). `--verbose` makes no difference.

## What the trace shows

`--log-level trace` with three file arguments prints the full config-load block three times, once before each `scanning path path=…` line: `using config from env var`, `extending config with default config`, `translated global prefilter expression`, `translated global filter expression`, and one `translated rule filter expression` line per default rule (several hundred lines per path). So every path argument re-reads the TOML, re-extends it with the default rule set, and re-translates every filter expression before scanning that one file.

## Why it matters

Pre-commit and CI hook runners such as `hk` pass the selected files as arguments (the hk builtin runs `betterleaks dir --redact --verbose --no-banner {{files}}`), and a `--all` run hands over every tracked file. In our GitHub Actions job that step takes 100–166 s; `dir .` over the same checkout takes about a second.

## Expected

Load and compile the config once per invocation and reuse it across path arguments. If per-target config discovery (issues #85 / #87) is the reason each path resolves its config separately, a cache keyed by the resolved config path would keep that behaviour while removing the repeated parse.

lekanjava:
## Summary

Fixes #343.

`betterleaks dir` previously reloaded and recompiled the same configuration once for every path argument. This made scans with thousands of explicit files much slower than scanning the same files through a directory path.

Add an invocation-scoped cache keyed by the effective configuration source. This preserves per-directory config discovery while reusing identical configs during one directory scan.

## Testing

- `go test ./cmd ./config`
- `git diff --check`

### DIFF
--- a/cmd/directory.go
+++ b/cmd/directory.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"github.com/betterleaks/betterleaks/config"
 	"github.com/betterleaks/betterleaks/detect"
 	"github.com/betterleaks/betterleaks/logging"
 	"github.com/betterleaks/betterleaks/sources"
@@ -26,6 +27,11 @@ var directoryCmd = &cobra.Command{
 }
 
 func runDirectory(cmd *cobra.Command, args []string) {
+	// A directory scan can receive thousands of individual file paths. Keep
+	// config discovery per source, but reuse each resolved config for the
+	// duration of this invocation.
+	configCache = make(map[string]*config.Config)
+
 	sourcesList := args
 	if len(sourcesList) == 0 {
 		sourcesList = []string{"."}
--- a/cmd/directory.go
+++ b/cmd/directory.go
@@ -31,6 +31,7 @@ func runDirectory(cmd *cobra.Command, args []string) {
 	// config discovery per source, but reuse each resolved config for the
 	// duration of this invocation.
 	configCache = make(map[string]*config.Config)
+	defer func() { configCache = nil }()
 
 	sourcesList := args
 	if len(sourcesList) == 0 {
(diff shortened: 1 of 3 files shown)


lekanjava:
Addressed the Greptile review: the config cache is now initialized and cleared only within `runDirectory`. When no invocation-scoped cache is active, `cachedConfig` loads directly without retaining the result. Added a regression test covering both reuse during a scan and no persistence outside the invocation. Focused tests pass: `go test ./cmd ./config`.

PR #349, opened 2026-09-10
Return an empty array for JSON reports without findings

### LINKED ISSUE #339: JSON report outputs null instead of [] when no findings are detected
When BetterLeaks generates a JSON report and no secrets/findings are detected, the output file contains:

JSON
null

instead of an empty JSON array:

JSON
[]

Since the JSON report represents a collection of findings, I would expect the output type to remain consistent regardless of the number of findings.

For example:

* Multiple findings → [{...}, {...}]
* One finding → [{...}]
* No findings → []

Currently, the last case returns null.

Expected behavior

When a scan completes successfully and no findings are detected, the JSON report should contain:

JSON
[]

Actual behavior

The generated JSON report contains:

JSON
null

Why this matters

Returning an empty array makes the JSON output easier and safer to consume programmatically, especially in CI/CD pipelines.

It also preserves the type of the report: consumers can always expect an array of findings rather than having to handle both an array and null.

For comparison, Gitleaks outputs an empty array ([]) when no secrets are detected.

Possible cause

This may be caused by a nil Go slice being passed to json.Encoder.

In Go:

var findings []Finding

is serialized by encoding/json as:

null

while an initialized empty slice:

findings := []Finding{}

is serialized as:

[]

The JSON reporter appears to encode the provided findings slice directly, so ensuring that an empty findings collection is initialized rather than nil (or normalizing it in the reporter) could resolve the issue.

Suggested behavior

A successful scan should always produce a JSON array:

Findings detected     → [{...}, {...}]
No findings detected  → []
Scan/reporting error   → non-zero exit code / error

danielgaskins:
## Summary

- Normalize nil findings before JSON encoding so successful reports always contain an array
- Cover nil findings separately from an initialized empty slice

Closes #339

## Testing

- `go test -race -p 2 ./...`
- `go build -p 2 ./...`
- `go run -p 2 . config check config/betterleaks.toml`
- `go run -p 2 . --regex-engine stdlib config check config/betterleaks.toml`
- `git diff --check`

### DIFF
--- a/report/json.go
+++ b/report/json.go
@@ -11,6 +11,9 @@ type JsonReporter struct {
 var _ Reporter = (*JsonReporter)(nil)
 
 func (t *JsonReporter) Write(w io.WriteCloser, findings []Finding) error {
+	if findings == nil {
+		findings = []Finding{}
+	}
 	encoder := json.NewEncoder(w)
 	encoder.SetIndent("", " ")
 	return encoder.Encode(findings)
--- a/report/json_test.go
+++ b/report/json_test.go
@@ -47,6 +47,11 @@ func TestWriteJSON(t *testing.T) {
 			testReportName: "empty",
 			expected:       filepath.Join(expectPath, "report", "empty.json"),
 			findings:       []Finding{}},
+		{
+			testReportName: "nil",
+			expected:       filepath.Join(expectPath, "report", "empty.json"),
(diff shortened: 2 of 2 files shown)


rgmz (approved):
Looks like a straightforward UX fix!

PR #353, opened 2026-09-12
fix: use Git-compatible null config path on Windows

### LINKED ISSUE #352: fix: Git history scans fail on Windows when config isolation uses NUL
### Summary

`betterleaks git <repo>` fails on current Git for Windows when Betterleaks isolates Git configuration with `GIT_CONFIG_GLOBAL=NUL` and `GIT_CONFIG_SYSTEM=NUL`.

### Reproduction

On Windows 11 with Git for Windows 2.53:

```powershell
$env:GIT_CONFIG_NOSYSTEM = '1'
$env:GIT_CONFIG_GLOBAL = 'NUL'
$env:GIT_CONFIG_SYSTEM = 'NUL'
git -C $repo log -1
```

Git exits 128:

```text
fatal: unable to access 'NUL': Invalid argument
```

The same failure surfaces from Betterleaks as `Failed to scan Git repository`. Filesystem scanning still succeeds, but Git history is left unscanned.

### Cause

`sources/git.go` and `sources/scm/clone.go` explicitly use `NUL` for the two config paths on Windows. Git for Windows 2.53 no longer accepts the device name as a config-file path.

### Expected behavior

Git configuration remains isolated without relying on the incompatible Windows device path. A Windows test should execute a real Git command with the generated environment so this cannot regress.

### Proposed fix

Use the Git-for-Windows-compatible `/dev/null` path on Windows (while retaining `GIT_CONFIG_NOSYSTEM=1`) or a lifecycle-managed empty regular file, and cover both history commands and clone commands.

danielwaldow-g2i:
Git history scans fail before reading any commits on Git for Windows 2.53 because the isolated child environment points both Git config files at `NUL`, which Git now rejects as a config path.

Use `/dev/null`, which Git for Windows maps correctly, for history and clone subprocesses on every platform. The new integration test runs a real Git command under the isolated environment, so Windows CI covers the reported failure rather than only comparing environment strings.

Closes #352.

Validation:

- `git diff --check`
- Full Linux and Windows test matrices run in CI

### DIFF
--- a/sources/git.go
+++ b/sources/git.go
@@ -13,7 +13,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
-	"runtime"
 	"strings"
 	"time"
 	"unicode"
@@ -38,12 +37,9 @@ type GitCmd struct {
 // gitConfigIsolationEnv contains the standard Git configuration isolation environment variables.
 // These settings prevent Git from reading user or system configuration files.
 func gitConfigIsolationEnv() []string {
-	var nullDevice string
-	if runtime.GOOS == "windows" {
-		nullDevice = "NUL"
-	} else {
-		nullDevice = "/dev/null"
-	}
+	// Git for Windows accepts its POSIX null path, while recent releases reject
+	// the Win32 NUL device name when it is used as a config-file path.
+	nullDevice := "/dev/null"
 	overrides := map[string]string{
 		"GIT_CONFIG_GLOBAL":      nullDevice,
 		"GIT_CONFIG_NOSYSTEM":    "1",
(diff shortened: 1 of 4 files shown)


danielwaldow-g2i:
Windows regression validation passed against this PR commit:

- Windows runner: `go test ./sources ./sources/scm` passed, including the real Git command under `gitConfigIsolationEnv`.
- Ubuntu runner: the same package tests passed.
- Windows x86_64 binary built successfully from the tested revision.

Evidence: https://github.com/danielwaldow-g2i/betterleaks/actions/runs/34705994162

The upstream pull-request workflow is waiting for a maintainer to authorize execution because this PR comes from a fork.

